Issue #27, opened 2021-11-04
GitHub workflow for backend linting


BerinikeTech:
## User story
1. As a developer
2. I want a GitHub workflow that automatically lints the (Python) backend code on every push
3. So that uniform coding style is enforced

## Acceptance criteria
* GitHub workflow is created
* Workflow is tested with one of the implemented backend functionalities

steinkohl:
Implementation also includes:
- Bandit security linting (avoid dangerous code patterns)
- Black formatter (only checks if correct format is given, actual formatting is done locally via pre-commit hook)
- Flake8 linter (checks code complexity)
- Pytest (tests python code and creates a ticket/issue if the required coverage isn't satisfied or a test fails)

PR #38, opened 2021-11-10
27 backend workflow

steinkohl:
Implementation of a backend ci pipeline according #27


.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
```
dominic-heil, line 1: I think a small Readme (or a small section in the main Readme) would be nice on how to actually utilize this file.

.github/workflows/python_backend_lint.yml
```diff
@@ -0,0 +1,78 @@
+name: Lint and Test Python Backend
+
+on: [push, pull_request]
```
dominic-heil, line 3:
```suggestion
on: [push]
```

On push should be sufficient I think, no? If you create a new PR, then the last push will have the workflows run already and this would only waste computing power.

dominic-heil (commented):
Well done. Two small comments but the rest LGTM!

dominic-heil:
Also, I think it would be nicer to have the linting and the tests split into two workflows so that you one can see quicker which workflow exactly failed. What do you think?

backend/python_service/service.py
```diff
@@ -0,0 +1,2 @@
+def hello_world():
```
dominic-heil, line 1: The copyright header is missing.

backend/python_service/test/test_service.py
```diff
@@ -0,0 +1,11 @@
+import unittest
```
dominic-heil, line 1: The copyright header is missing.

.pre-commit-config.yaml
steinkohl: Thats a point, I will put in a reference to the official manual

.github/workflows/python_backend_lint.yml
steinkohl: Well, for testing it actually makes sense, since the tests then will perform on the files how the would be in the merged state.
But I will split the workflow apart, as you suggested, then only the test part will be performed again

backend/python_service/service.py
steinkohl: fixed

backend/python_service/test/test_service.py
steinkohl: fixed

.pre-commit-config.yaml
steinkohl: fixed

.github/workflows/python_backend_lint.yml
dominic-heil: Good point!

dominic-heil (commented):
LGTM

PR #41, opened 2021-11-10
attempt to only eslint frontend and not any other files

dominic-heil:
This PR implements a fix so that only the frontend directory is linted and not any other JS files.


BerinikeTech (approved):
As requested

PR #43, opened 2021-11-10
Connect wallet to account

tuilagio:
- Feature: user can connect with MetaMask plugin. After that, fetch other information about this public address such as Balance, ...
   

- Improve: add instruction to start the Appwrite backend and setup frontend in `frontend/README.md` 
- Change: move `REACT_APP_ENDPOINT` and `REACT_APP_PROJECT` to `frontend/.env`. Add the environment file to ignore list.
- TODO: for future work: support more wallets: Coinbase, ... Add style.


frontend/src/components/CenterFlexBoxMedium.js
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2021 Dominic Heil <d.heil@campus.tu-berlin.de>
```
dominic-heil, line 2: I did not author this file. Please change this to your information (also for the other files).

frontend/README.md
```diff
@@ -1,3 +1,30 @@
...
+    REACT_APP_PROJECT="618a936863851"
+    ```
+    Additionally, these data can also be written in `/frontend/src/utils/config.js`.
+    Seem to be that API key is not needed. Now the web UI can be used.
+
+#
+```bash
+# Make 
+```
+
```
dominic-heil, line 27: Well written!

frontend/src/components/Header.js
```diff
@@ -34,6 +34,7 @@ export default function Header({ children, user }) {
 						<Typography variant="h6" component="div" sx={{ mr: 2, flexGrow: 1 }} onClick={() => routeChange("/faqs")}  style={{ cursor: "pointer" }}>
 							FAQS
 						</Typography>
+						<Button color="inherit" component={Link} to="/wallets">Wallets</Button>
```
dominic-heil, line 37: This button should only be visible if the user is logged in. Right now it throws an exception if the user is not logged in when the user clicks this button. I think later we will probably move this into the profile and thus it should not be a problem, but for now maybe do something similar as a few lines below, i.e. check if user is null and return <></> if this is the case or otherwise return the wallets button?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+
+	useEffect(() => {
+		async function fetchAccountsDetails(publicAddresses) {
+			const ethereum = await detectEthereumProvider();
+			if (!ethereum) {
+				console.error("no ethereum");
+				return;
+			}
+			let data = [];
+			// This is the most bullshit thing I've seen. Can't use ForEach nor For...in!
```
dominic-heil, line 38: Read this if you want to understand why: https://stackoverflow.com/a/7190438/7838440

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+				let pA = publicAddresses[i];
+				const bl = await ethereum.request({ method: "eth_getBalance", params: [pA, "latest"] });
+				// console.log("eth_getBalance: " + bl + " (" + parseInt(bl, 16) + " Ether)");
+				if (bl) {
+					console.log("bl=" + bl);
+					data.push({ 
+						pubAddres: pA,
+						bl: bl
+					});
+				}
```
dominic-heil, line 50: I think a more descriptive variable, e.g. `balance`, would fit better here. The same for the other places that use `bl`.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			spacing={2}
+			alignItems="center"
+			justifyContent="center"
+			direction="column">
+			<Grid item style={{ width: "100%" }}>
+				{!publicAddresses
+					?
+					<Button color="inherit" onClick={handleAddMetaMask}>Connect MetaMask wallet</Button>
+					:
+					<EthAccountsDetail publicAddresses={publicAddresses}></EthAccountsDetail>
```
dominic-heil, line 138:
```suggestion
					<EthAccountsDetail publicAddresses={publicAddresses}/>
```

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+					data.push({ 
+						pubAddres: pA,
+						bl: bl
+					});
+				}
+			}
+			setDetailData(data);
+			// console.log({ detailData: detailData });
+		}
+		fetchAccountsDetails(publicAddresses);
```
dominic-heil, line 55: For a better structure and understandability, I would probably only call the function here and define the function above the component.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+import { useHistory } from "react-router-dom";
+import Grid from "@mui/material/Grid";
+import { useState } from "react";
+import detectEthereumProvider from "@metamask/detect-provider";
+import CenterFlexBoxMedium from "../components/CenterFlexBoxMedium";
+
+function AccountEntry({ data }) {
+	return 	<TableRow>
+		<TableCell style={{ color: "white", borderBottom: "none" }}>{data.pubAddres}</TableCell>
+		<TableCell style={{ color: "white", borderBottom: "none" }}>{parseInt(data.bl, 16)}</TableCell>
```
dominic-heil, line 23: I do not think this is an int. I only 0.004950511454860081 Ether. 
image

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+
+	return <TableContainer  style={{ color: "white" }}>
+		<Table>
+			<TableBody>
+				<TableRow>
+					<TableCell style={{ color: "white", borderBottom: "none" }}>Account (Public Address)</TableCell>
+					<TableCell style={{ color: "white", borderBottom: "none" }}>Balance (Eth)</TableCell>
+				</TableRow>
+				{detailData.map((value, key) => {
+					return <AccountEntry key={key} data={value}></AccountEntry>;
```
dominic-heil, line 66:
```suggestion
					return <AccountEntry key={key} data={value}/>;
```

frontend/src/App.js
```diff
@@ -15,6 +15,7 @@ import EmailConfirmPage from "./pages/EmailConfirmPage";
 import RequestPasswordResetPage from "./pages/RequestPasswordResetPage";
```
dominic-heil, line 15: As you edited this file, I think you should add yourself to the copyright at the top of the file.

dominic-heil:
Functionality wise, most things work pretty well. Good job!

frontend/src/pages/Wallet.js
dominic-heil: Maybe also add some (debug) error handling like .catch(e => {console.log(e);}) ?

frontend/src/pages/Wallet.js
krisutofu, line 23: Could we use sensible naming here? `balance` instead of `bl`? This is not a dirty C school project :-) .

dominic-heil: I already commented the same thing above :)

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			const ethereum = await detectEthereumProvider();
+			if (!ethereum) {
+				console.error("no ethereum");
+				return;
+			}
+			let data = [];
+			// This is the most bullshit thing I've seen. Can't use ForEach nor For...in!
+			const length = publicAddresses.length;
+			for (let i=0; i<length; i++) {
+				let pA = publicAddresses[i];
```
krisutofu, line 41: Please don't use acronym variable names. This is not dramatic in this case, but good variable naming practice is if names can be spoken out loud and allow other people to understand what is talked about.

dominic-heil: Do you mean the variable `i`?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+				const bl = await ethereum.request({ method: "eth_getBalance", params: [pA, "latest"] });
+				// console.log("eth_getBalance: " + bl + " (" + parseInt(bl, 16) + " Ether)");
+				if (bl) {
```
krisutofu, line 44: Could you make this predicate more concrete as a boolean expression? The intension behind using a number as boolean is not clear. Is this a check for NaN, 0 or both? Why excluding 0?

Could you use a self-describing name which explains the condition? Like `const isValidBalance = balance > 0`.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+		<TableCell style={{ color: "white", borderBottom: "none" }}>{parseInt(data.bl, 16)}</TableCell>
+	</TableRow>;
+}
+
+function EthAccountsDetail({ publicAddresses }) {
+	const [detailData, setDetailData] = useState([]);
+
+	useEffect(() => {
+		async function fetchAccountsDetails(publicAddresses) {
+			const ethereum = await detectEthereumProvider();
```
krisutofu, line 32: Could you use a boolean-indicating name like `hasDectedEthereumProvider` instead of `ethereum`?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			if (!ethereum) {
+				console.error("no ethereum");
+				return;
+			}
+			let data = [];
```
krisutofu, line 37: I think `data` is a bad variable name in general. I don't see any types here, so I would suggest encoding the type into the variable name like `walletRecords` or `walletList`.

frontend/src/pages/Wallet.js
krisutofu, line 38: This comment should be removed from the merged version. I hope you understand.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			// This is the most bullshit thing I've seen. Can't use ForEach nor For...in!
+			const length = publicAddresses.length;
+			for (let i=0; i<length; i++) {
+				let pA = publicAddresses[i];
+				const bl = await ethereum.request({ method: "eth_getBalance", params: [pA, "latest"] });
+				// console.log("eth_getBalance: " + bl + " (" + parseInt(bl, 16) + " Ether)");
```
krisutofu, line 43: Please _don't_ submit commented code. I hope there is some kind of debug-logger for JavaScript?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+				</TableRow>
+				{detailData.map((value, key) => {
+					return <AccountEntry key={key} data={value}></AccountEntry>;
+				})}
+			</TableBody>
+		</Table>
+	</TableContainer>;
+}
+
+export default function Wallet({ user, setUser }) {
```
krisutofu, line 73: Would it make sense to create a `class` from this function?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+	const [publicAddresses, setPublicAddresses] = useState("");
+	const history = useHistory();
+
+	const routeChange = (path) =>{
```
krisutofu, line 77: This code is cloned so often already. Could this be put in a base class? I am wondering if someone wouldn't already made a class with such a function?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+		history.push(path);
+	};
+
+	useEffect(() => {
+		appwriteApi.getAccount()
+			.then(res => setUser(res));
+	}, []);
+
+	if (!user){
```
krisutofu, line 86: What is `user`? A global variable and why? This name does not sound like a boolean so the intension is not clear. Should this be a `null` check? You can add a self-describing name like `const hasLoggedInUser = (user != null)`.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+		routeChange("/");
+		return <></>;
+	}
+
+	/* TODO: Fetch public address for this user from backend */
```
krisutofu, line 91: Is this still todo? If no, this comment should be removed.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+
+	/* Handle button connect to MetaMask */
+	const handleAddMetaMask = async () => {
+		const ethereum = await detectEthereumProvider();
+		// Ref: https://docs.metamask.io/guide/rpc-api.html#table-of-contents
+		let pubAddr = 0;
+		if (ethereum) {
```
krisutofu, line 98: Could you make use of the "Early-Exit" pattern?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
+			const account = accounts[0];
+			pubAddr = account;
+			if (publicAddresses) {
+				setPublicAddresses(publicAddresses.push(pubAddr));
```
krisutofu, line 103: Is `publicAddresses` global? Would it make more sense to write `walletAddress` instead of `Address`? Is a wallet not just an address so that we can say even `wallet`?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+			console.log("Please install MetaMask!");
+			return;
+		}
+
+		/* Some playground code 
+		* A complete list of requests can be found here
+		* https://eth.wiki/json-rpc/API
+		*/
+		// // Get gas Price
+		// const gp = await ethereum.request({ method: "eth_gasPrice", params: [] });
```
krisutofu, line 117: Is there some kind of debug-code-block in JavaScript where this could be logged instead of commented?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+
+	return <CenterFlexBoxMedium style={{ maxWidth: "md" }}>
+		<Grid
+			container
+			spacing={2}
+			alignItems="center"
+			justifyContent="center"
+			direction="column">
+			<Grid item style={{ width: "100%" }}>
+				{!publicAddresses
```
krisutofu, line 134: I think, the condition should get a better name which makes the intension clear and looks like a condition rather than something else.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,143 @@
...
+					?
+					<Button color="inherit" onClick={handleAddMetaMask}>Connect MetaMask wallet</Button>
+					:
+					<EthAccountsDetail publicAddresses={publicAddresses}></EthAccountsDetail>
+				}
+			</Grid>
+		</Grid>
+	</CenterFlexBoxMedium>;
+}
```
krisutofu, line 143: Well done Que!

frontend/src/components/Header.js
```diff
@@ -34,6 +34,7 @@ export default function Header({ children, user }) {
...
 						{ user
```
krisutofu, line 38: Would it be possible to use a intensional boolean predicate name (like in `Wallet.js`) instead of `user`?

frontend/src/pages/Wallet.js
krisutofu: More like variable `pA` and `bl`. `i` is one of the exceptions where I don't care about, because that's idiomatic naming ;-) .

frontend/src/pages/Wallet.js
krisutofu: Yeah, you did :-D . This was the first thing I saw and needed to comment.

frontend/src/pages/Wallet.js
tuilagio: no, because an Ethereum-wise object is returned if success

frontend/src/pages/Wallet.js
tuilagio: I don't know if that is possible/best practice. May be decide later in Refactoring sprint?

frontend/src/pages/Wallet.js
tuilagio: todo because I am waiting for the Back end to provide my necessary endpoints

frontend/src/pages/Wallet.js
tuilagio: There is wallet that manages addresses. This state holds multiple addresses

frontend/src/pages/Wallet.js
tuilagio: error handling already coded in the function

frontend/src/pages/Wallet.js
dominic-heil: I implemented a custom React hook to do that recently. We can change that then I think.

frontend/src/pages/Wallet.js
krisutofu: I thought about an additional variable `const hasDectectedEthereumProvider =  (ethereum != null)` in between.

In this case, it's not dramatic, you could also inline it. But using an arbitrary variable as boolean value which without  a name that would indicate boolean meaning is not generally clear in my opinion.

PR #52, opened 2021-11-11
added mockups for fronted

NickZecamp:


dominic-heil:
@NickZecamp We talked with Andre about the mock ups today. He liked the general design and we can start implementing the design in the next sprint. So good job! The only change that I would request is that I think the deliverables directory is not the best directory for the figma file. Maybe also put it in a `assets` folder or something else?

tuilagio:
Just for the next time. May be a few screenshot would be nicer, because then anyone can see the GUI directly here.

PR #54, opened 2021-11-13
init backend-connect-wallet

Floinko:


steinkohl (commented):
Looks good to me!
+ nice readmes
+ well structured code
- maybe add some more comments to the code? but I think it's quite self-explanatory

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -0,0 +1,39 @@
...
+
+desiredWalletAddress = json.loads(PAYLOAD)["walletAddress"]
+
+readResult = database.list_documents(WALLETS_COLLECTION_ID, ["userId=" + USER_ID])["documents"]
+
+if len(readResult) > 0:
+    writeResult = database.update_document(WALLETS_COLLECTION_ID, readResult[0]["$id"],
+                                           {"walletAddress": desiredWalletAddress})
+else:
+    writeResult = database.create_document(WALLETS_COLLECTION_ID, {"userId": USER_ID, "walletAddress": desiredWalletAddress})
```
dominic-heil, line 33: I think we should also allow admins to read this document, no?

dominic-heil:
Also, didn't we say that we only need update/create functions and that we would use the already existing. protected appwrite API endpoints for retrieval? The extra get function feels like a little bit of extra overhead to me.

backend/appwrite-functions/set-wallet/main.py
Floinko: AFAIK admins are also users

dominic-heil: But they also belong to the admin team afaik (based on the other scripts where we add some admins). To the best of my knowledge, if you only set one user, then only that user can read it

Floinko:
> Also, didn't we say that we only need update/create functions and that we would use the already existing. protected appwrite API endpoints for retrieval? The extra get function feels like a little bit of extra overhead to me.

Yeah i get what you mean.  On the other hand it simplifies the procedure for the frontend since only the functions API is used. Also it is already implemented and works fine like this. Would you be fine with leaving this like it is and keeping this other approach in mind for future features?

backend/appwrite-functions/set-wallet/main.py
Floinko: Why should admins be able to see any users address?

dominic-heil:
> > Also, didn't we say that we only need update/create functions and that we would use the already existing. protected appwrite API endpoints for retrieval? The extra get function feels like a little bit of extra overhead to me.
> 
> Yeah i get what you mean. On the other hand it simplifies the procedure for the frontend since only the functions API is used. Also it is already implemented and works fine like this. Would you be fine with leaving this like it is and keeping this other approach in mind for future features?

For consistency, this would mean that every get endpoint is implemented in this way (instead of using the appwrite API for that) and the error handling needs to be done on in the frontend again anyways. I wouldn't mind in the frontend, but I think this adds extra work for the backend in the future.

backend/appwrite-functions/set-wallet/main.py
dominic-heil: I have no use case for it now, but shouldn't admins have access to general user data? If you do not think there is a use case for it, leave admins out

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -0,0 +1,39 @@
...
+
+(client
+ .set_endpoint(APPWRITE_ENDPOINT)
+ .set_project(APPWRITE_PROJECT)
+ .set_key(APPWRITE_API_KEY)
+ )
+
+desiredWalletAddress = json.loads(PAYLOAD)["walletAddress"]
+
+readResult = database.list_documents(WALLETS_COLLECTION_ID, ["userId=" + USER_ID])["documents"]
```
krisutofu, line 27: `readResult` is kind of generic name which doesn't me as reader about the content. Looks like a list of something. What does it contain?

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -0,0 +1,39 @@
...
+
+# provided by cloud function env vars
+APPWRITE_ENDPOINT = os.environ.get('APPWRITE_ENDPOINT')
+APPWRITE_API_KEY = os.environ.get('APPWRITE_API_KEY')
+WALLETS_COLLECTION_ID = os.environ.get('WALLETS_COLLECTION_ID')
+
+client = Client()
+database = Database(client)
+
+(client
```
krisutofu, line 19: Are these settings related to the database query? Would it make a difference if these setters are written before the `Database` constructor?

backend/database-collection-schemas/createWalletsCollection.py
```diff
@@ -0,0 +1,36 @@
...
+database = Database(client)
+
+(client
+    .set_endpoint(APPWRITE_ENDPOINT)
+    .set_project(APPWRITE_PROJECT)
+    .set_key(APPWRITE_API_KEY)
+ )
+
+createCollectionResult = database.create_collection('Wallets', [], [], [
+    {
```
krisutofu, line 20: would it make sense to create a datat structure with constructor for these dictionaries?

backend/appwrite-functions/set-wallet/main.py
Floinko: As long as we don't have a use case for that i would restrict access as far as possible

backend/appwrite-functions/set-wallet/main.py
Floinko: It queries the list of documents containing wallet addresses for the requesting user. The expected amount of documents is either 0 or 1. Since there is only one read operation in the whole file i think it is descriptive enough. Might add more comments tho.

backend/appwrite-functions/set-wallet/main.py
Floinko: Don't know which setters you mean.

backend/database-collection-schemas/createWalletsCollection.py
Floinko: I think its very readable that way. I also don't know how a constructor for those setting would make the code shorter.

Floinko:
> > > Also, didn't we say that we only need update/create functions and that we would use the already existing. protected appwrite API endpoints for retrieval? The extra get function feels like a little bit of extra overhead to me.
> > 
> > 
> > Yeah i get what you mean. On the other hand it simplifies the procedure for the frontend since only the functions API is used. Also it is already implemented and works fine like this. Would you be fine with leaving this like it is and keeping this other approach in mind for future features?
> 
> For consistency, this would mean that every get endpoint is implemented in this way (instead of using the appwrite API for that) and the error handling needs to be done on in the frontend again anyways. I wouldn't mind in the frontend, but I think this adds extra work for the backend in the future.

Added the necessary permissions to the collection and individual documents, so that users can retrieve their wallet addresses directly. Deleted the get-wallet-address appwrite-function. Much faster and less complicated now :+1:

steinkohl:
> Looks good to me! 
> ++ nice readmes 
> ++ well structured code -- maybe add some more comments to the code? but I think it's quite self-explanatory

sanitation for the wallet data also would be useful

dominic-heil (approved):
LGTM! NIce, self explanatory code!

Floinko:
> Looks good to me! ++ nice readmes ++ well structured code -- maybe add some more comments to the code? but I think it's quite self-explanatory

added comments

Floinko:
> > Looks good to me!
> > ++ nice readmes
> > ++ well structured code -- maybe add some more comments to the code? but I think it's quite self-explanatory
> 
> sanitation for the wallet data also would be useful

added sanitation

backend/appwrite-functions/set-wallet/main.py
krisutofu: Okay but please no further comments. Your code is well written and I can understand it well except for this variable.

Variable names are supposed to provide additional clarity. For improved clarity you could rename the variable to `walletDocumentList` or `listOfWalletDocuments` or `walletDocumentsOfUser` or ...

If there are explicit expectations, the best way to make them transparent and safe are `assert`ions. This is even more useful because Python is naturally intransparent with types, constraints or expectations. Assumptions are unsafe – `assume` exists in verification theory which is used for debugging only (they add knowledge at a certain point without proving them).
Contracts are almost like types but unlike types they only hold at specific points of time.
I would expect Python to have `assert`s like any modern programming language which feature at least a subset of design-by-contract programming. A safe programming language could be able to prove asserts.

Comments only make sense for meta communication or inline documentation for public symbols, mainly for reasons why something was programmed in a certain way or for learning purposes. Learning purposes make sense for open-source projects which can be accessed by beginners or unfamiliar people. I also used comments to remember cryptic compiler errors that I solved.

backend/appwrite-functions/set-wallet/main.py
krisutofu: Okay, I mean: would it make a difference to write this `client` statement in parentheses with its `client.setX` calls before the database?

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -8,38 +10,47 @@
 PAYLOAD = os.environ.get("APPWRITE_FUNCTION_DATA")
 APPWRITE_PROJECT = os.environ.get("APPWRITE_FUNCTION_PROJECT_ID")
 
-# provided by cloud function env vars
+# provided by appwrite function env vars
 APPWRITE_ENDPOINT = os.environ.get("APPWRITE_ENDPOINT")
 APPWRITE_API_KEY = os.environ.get("APPWRITE_API_KEY")
 WALLETS_COLLECTION_ID = os.environ.get("WALLETS_COLLECTION_ID")
 
+# instantiate database connection
```
krisutofu, line 18: I see the good intent but personally, I find comments like these counter-productive. Same for the added comments below. Of course, depends on your decision, but I wouldn't do that.

The professional software-engineering way of explaining code is by outlining code with good function names. My personal experience is that additional comments disturb the readability, particularly if the code is clear, but outlining them into functions is better (more levels of detail).

The function would be called  `instantiateDatabaseConnection()`.

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -8,38 +10,47 @@
...
+client.set_key(APPWRITE_API_KEY)
 database = Database(client)
 
-(
-    client.set_endpoint(APPWRITE_ENDPOINT)
-    .set_project(APPWRITE_PROJECT)
-    .set_key(APPWRITE_API_KEY)
-)
-
+# parse wallet address
```
krisutofu, line 25: How is this:
`desiredWalletAddress = parseWalletAddress()`

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -8,38 +10,47 @@
...
 desiredWalletAddress = json.loads(PAYLOAD)["walletAddress"]
 
+# validate that it is an ethereum address
```
krisutofu, line 28: `isValidEthereumAddress(desiredWalletAddress)`

backend/appwrite-functions/set-wallet/main.py
```diff
@@ -8,38 +10,47 @@
...
     writeResult = database.create_document(
         WALLETS_COLLECTION_ID,
         {"userId": USER_ID, "walletAddress": desiredWalletAddress},
         ["user:" + USER_ID],
     )
 
-res = {"status": "success", "dbResult": writeResult}
+# return that the operation was successful
+res = {"status": "success"}
 print(json.dumps(res))
```
krisutofu, line 56: Would this be good:

```
instantiateDatabaseConnection()

desiredWalletAddress = parseWalletAddress()

checkValidEthereumAddress(desiredWalletAddress)

usersWalletDocument = readDatabaseWalletDocument(USER_ID)

writeStatus = writeDatabaseWalletDocument(usersWalletDocument, desiredWalletAddress)

printOperationStatus(writeStatus)
```

Only a suggestion for structuring it, if oneone thinks, it is not clear enough.

backend/database-collection-schemas/createWalletsCollection.py
krisutofu: Okay. At the current state, it makes not a big difference. Only a recommendation.

My thought was about maintenance, i.e. the cost of change or adding new functions later which could use this dictionary structure. New functions could need to create such objects. I imagined defining a simple datastructure in the good old OOP way, which allows structures to be reusable and easier to change because the structure is defined in a single location.

Tuples and dictionary are good for code which doesn't need to scale or be changed often because they are one-time objects without reusable structure. Reduced scalability can be seen because it requires you to code-clone the dictionary structure a 2nd time.

For quality, I think, shortness of code is mostly irrelevant. Compressing code often counters maintenance because the code is more specifically written for a specific feature (code is coupled to a specific feature).

PR #55, opened 2021-11-13
47 dropdate nft

JaPilgrim:
-Added Droptime blocker for buying function
-Configured for Kovan Testnet Deployment with truffles & Infura


blockchain/contracts/HelloWorld.sol
```diff
@@ -25,15 +27,18 @@ contract HelloWorld {
         NFTOwnership memory nftOwnership;
         nftOwnership.nftId = nftHash;
         nftOwnership.owner = nftOwner;
+        nftOwnership.droptime = 1637085033;
```
BerinikeTech, line 30: This should not be hardcoded but rather be a parameter of the mockNFT function

blockchain/contracts/HelloWorld.sol
```diff
@@ -25,15 +27,18 @@ contract HelloWorld {
...
         nftOwnerships[nftHash] = nftOwnership;
         nfts.push(nftHash);
         return nftHash;
     }
 
-    function buyNFT(uint256 price, uint256 nftHash) public payable {
+    function buyNFT(uint256 price, uint256 nftHash) public payable returns(address) {
+        require(nftOwnerships[nftHash].droptime<=block.timestamp,string(abi.encode("Drop has not started yet! ",(nftOwnerships[nftHash].droptime-block.timestamp),"Seconds left!")));
```
BerinikeTech, line 37: I think we should change that to not display seconds but hours:minutes:seconds

blockchain/contracts/HelloWorld.sol
```diff
@@ -25,15 +27,18 @@ contract HelloWorld {
...
         price = msg.value;
         nftOwner.transfer(price);
         nftOwnerships[nftHash].owner = buyer;
+        return nftOwnerships[nftHash].owner;
```
BerinikeTech, line 41: Why did you add this return value?

BerinikeTech (changes requested):

blockchain/contracts/HelloWorld.sol
JaPilgrim: Good idea!

blockchain/contracts/HelloWorld.sol
JaPilgrim: Thought it would be beneficial for testing but turns out its not, resolved

blockchain/contracts/HelloWorld.sol
JaPilgrim: Due to the fact solidity does not have built-in int->String parsing, nor string concatenation(https://ethereum.stackexchange.com/questions/10811/solidity-concatenate-uint-into-a-string), I would rather not implement any display of remaining time yet, as my approach didnt work anyways. Also the result would only be displayed in the blockchain-viewer like etherscan as far as I know & it should rather be implemented in the frontend in the future.

BerinikeTech: Fine with me :)

BerinikeTech (approved):
Changes were implemented as discussed

PR #56, opened 2021-11-17
45 change team of user

dominic-heil:
This PR implements a way to add users to the "Admins" team if the current user in the frontend is in the Admins team.

The admin can search for a user by their mail, where they can invite the user to the Admins team or remove them if they are part of the Admins team. 

image

To test this fully, appwrite must be able to send emails. Without a setup email, the invitation can be seen in the appwrite UI as "pending".


frontend/src/pages/EmailConfirmPage.js
```diff
@@ -19,9 +19,7 @@ export default function EmailConfirmPage() {
 			if (res){
 				setWasConfirmed(true);
 			}
-		}).catch(r =>
-			console.log(r)
-		);
+		}).catch(() => {});
```
Floinko, line 22: Why was this removed? Just ran into an error regarding the mail verification and had no information on that at all...

frontend/src/utils/utils.js
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+// SPDX-FileCopyrightText: 2021 Dominic Heil <d.heil@campus.tu-berlin.de>
+
+export const isValidEmail = (email) => {
+	const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
Floinko, line 5: Please add your source for this regex here.

Floinko (commented):
Code looks very good to me (except the two things i commented on) and everything worked for me when i tested it. However i think is still some work to do in terms of documentation:

- commenting in general
- how do i become the first admin? => create group "Admins" in appwrite
- which config do i have to change before starting the project? => currently harcoded for a local instance in config.js
- enable invites in appwrite

frontend/src/pages/EmailConfirmPage.js
dominic-heil: I added it back for now. I removed it as I thought we do not want to have anything written to the console when we deploy the frontend somewhere, but for now I think it makes sense to keep it in.

frontend/src/utils/utils.js
dominic-heil: Done.

dominic-heil:
@Floinko thank you for the feedback. I will add comments to each component/page later today, as well as update the readme in regards to setting up appwrite (i.e. linking to the backend scripts).

BerinikeTech (approved):
A little bit more documentation would definitely be helpful, but I think you did a great job naming the parameters, really straight forward and no room for misinterpretation

Floinko:
LGTM!

PR #58, opened 2021-11-17
added NFT images for Nuremberg

NickZecamp:


dominic-heil:
@NickZecamp Well done with the images!

I don't think that the images fit best in the deliverables folder. I would think if we create a new directory in the root directory called something like `assets` and add the images in a subdirectory of that directory, thay would be a good idea.

Also, do you have an image of the whole map of the sub images? That would be something useful for the website I think. It does not need to be in the original resolution of course.

NickZecamp:
> @NickZecamp Well done with the images!
> 
> I don't think that the images fit best in the deliverables folder. I would think if we create a new directory in the root directory called something like `assets` and add the images in a subdirectory of that directory, thay would be a good idea.
> 
> Also, do you have an image of the whole map of the sub images? That would be something useful for the website I think. It does not need to be in the original resolution of course.

Done!

PR #59, opened 2021-11-17
Updated init script

steinkohl:
adds the 'owner' role to 'Admins' team members


backend/python_init_script/main.py
```diff
@@ -0,0 +1,45 @@
+import os
```
dominic-heil, line 1: The copyright header is missing.

backend/python_init_script/main.py
```diff
@@ -0,0 +1,45 @@
...
+from appwrite.client import Client
+
+# from appwrite.services.database import Database
+from appwrite.services.users import Users
+from appwrite.services.teams import Teams
+
+# from appwrite.services.functions import Functions
```
dominic-heil, line 8: Can commented imports and statements (below) be removed?

dominic-heil (commented):
Tested it on my system with a new appwrite project and everything worked. Only have two small comments. Also, another thing that we might want to change: If there already is an "Admins" team, then a new Admins team is created anyways. Instead, if there already is an Admins team, then I think the previous Admins team should be user or the script should exit.

backend/python_init_script/main.py
steinkohl: stupid me again

backend/python_init_script/main.py
steinkohl: yes.

steinkohl:
> Tested it on my system with a new appwrite project and everything worked. Only have two small comments. Also, another thing that we might want to change: If there already is an "Admins" team, then a new Admins team is created anyways. Instead, if there already is an Admins team, then I think the previous Admins team should be used or the script should exit.

I will implement that in the next sprint including handling of user/admin data without exposing to public

dominic-heil (approved):

PR #61, opened 2021-11-17
Black update

steinkohl:
replaces cclauss github account with my account


tuilagio:
why this change is necessary?

steinkohl:
> why this change is necessary?

It's totally optional. Since you were confused that this user contributed to the repository, other might as well, so I thought it would be better if the contributor is one of us.

krisutofu (approved):
I am fine with this.

PR #76, opened 2021-11-18
Connect wallet to account

tuilagio:
- Feature: user can connect with MetaMask plugin. After that, fetch other information about this public address such as Balance, ...
   

- Improve: add instruction to start the Appwrite backend and setup frontend in `frontend/README.md` 
- Change: move `REACT_APP_ENDPOINT` and `REACT_APP_PROJECT` to `frontend/.env`. Add the environment file to ignore list.
- TODO: for future work: support more wallets: Coinbase, ... Add style.


frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+		<TableCell style={{ color: "white", borderBottom: "none" }}>
+			{data.pubAddress}
+		</TableCell>
+		<TableCell style={{ color: "white", borderBottom: "none" }}>
+			{utils.formatEther(data.balance)}
+		</TableCell>
+	</TableRow>;
+}
+
+function EthAccountsDetail({ publicAddresses }) {
```
krisutofu, line 33: I think, this could be a class which extends `React.Component`.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+import appwriteApi from "../api/appwriteApi";
+import { useHistory } from "react-router-dom";
+import Grid from "@mui/material/Grid";
+import { useState } from "react";
+import detectEthereumProvider from "@metamask/detect-provider";
+import CenterFlexBoxMedium from "../components/CenterFlexBoxMedium";
+const { utils } = require( "ethers" );
+// import utils from "ethers";
+
+function AccountEntry({ data }) {
```
krisutofu, line 22: Nice that you put this in its own function!

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+		</TableCell>
+	</TableRow>;
+}
+
+function EthAccountsDetail({ publicAddresses }) {
+	const [detailData, setDetailData] = useState([]);
+
+	async function fetchAccountsDetails(publicAddresses) {
+		const provider = await detectEthereumProvider();
+		if (!provider) {
```
krisutofu, line 38: It would be nice if boolean values / predicates would be used instead of implicit type casts. In cases where the condition is not clear, an additional variable name can help.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+			console.error("no ethereum provider");
```
krisutofu, line 39: For the design, will we need a user-facing error message later?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+			return;
+		}
+		let accDetail = [];
```
krisutofu, line 42: It would be nice if we can use variables that can be communicated well, and even understood by those who might not be even programmers. Abbreviations also have the problem that they can have multiple meanings like `accent`, `accident`, `acceleration`, `acceptance`, `accuracy`. I think, `acc` is not even a common abbreviation for `account`.

Another issue: a variable which contains a collection of things is singular here. I think a plural name would fit more like `accountDetails`. Even a better fit: `accountDetailList`.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+		const length = publicAddresses.length;
+		for (let i=0; i<length; i++) {
```
krisutofu, line 44: Could we turn this into a `publicAddresses.forEach()` ?

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
+			const account = accounts[0];
+			pubAddr = account;
+			if (publicAddresses) {
+				setPublicAddresses(publicAddresses.push(pubAddr));
+			} else {
+				setPublicAddresses([pubAddr]);
+			}
+		} else {
+			console.log("Please install MetaMask!");
```
krisutofu, line 114: This should be a user-facing message, not in the console.

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+	/* Handle button connect to MetaMask */
+	const handleAddMetaMask = async () => {
+		const ethereum = await detectEthereumProvider();
+		// Ref: https://docs.metamask.io/guide/rpc-api.html#table-of-contents
+		let pubAddr = 0;
+		if (ethereum) {
+			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
+			const account = accounts[0];
+			pubAddr = account;
+			if (publicAddresses) {
```
krisutofu, line 108:
```suggestion
			setPublicAddresses( (publicAddresses ?? []) .push (publicAddress) )
```
Also possible `publicAddresses ??= []` .

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	/* TODO: Fetch public address for this user from backend */
+
+	/* Handle button connect to MetaMask */
+	const handleAddMetaMask = async () => {
+		const ethereum = await detectEthereumProvider();
+		// Ref: https://docs.metamask.io/guide/rpc-api.html#table-of-contents
+		let pubAddr = 0;
+		if (ethereum) {
```
krisutofu, line 104: It would be better to use the "Early-Exit pattern" here.

```
if (ethereum == null) { ... return; }
const accounts = ...;
...
```

frontend/src/pages/Wallet.js
```diff
@@ -0,0 +1,149 @@
...
+			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
+			const account = accounts[0];
```
krisutofu, line 106: What is with the other accounts?

frontend/src/pages/Wallet.js
dominic-heil: See https://reactjs.org/docs/components-and-props.html. Either defining a component as as a function or as a class that extends React.Component is fine. I have seen the first approach used much more frequently.

dominic-heil: Also, as this is only used by the main component of the file, there is no need to put the small component in another file.

frontend/src/pages/Wallet.js
dominic-heil: `!provider` is just the short hand notation for `provider === null`. I think this is fine and common practice.

frontend/src/pages/Wallet.js
dominic-heil: I disagree with the acc part. First, code does not need to be understood by non programmers. Second, the variable is defined in the context of a function called `fetchAccountsDetails`, i.e. it is very obvious that acc does not mean accent, accident or anything else, but means account. I agree that it should be plural.

frontend/src/pages/Wallet.js
dominic-heil: I think it does not really matter if we use a for, for ... of or forEach loop.

frontend/src/pages/Wallet.js
dominic-heil: The "React way" of doing this would be something like this (not tested right now):

```javascript
setPublicAddresses(publicAddressesPrev => [...publicAddressesPrev, pubAddr])
```

And `const [publicAddresses, setPublicAddresses] = useState("");``should no be initialized as a String, if it is not a String, but as an empty list, i.e. []. In the code below (line 140), a length > 1 check instead of null check should be done.

dominic-heil (changes requested):
Please do not forget to add JDoc comments to large components. I also added a few comments to @krisutofu 's comments

frontend/src/pages/Wallet.js
dominic-heil: Instead of a console.log you could do something similar as I have done on other pages and define a errorMessage state that is initially empty, i.e. "", and if it is set to something non empty, it is displayer as an Alert:

```javascript
{errorMessage !== "" && <Grid item xs={12}><Alert severity="error">{errorMessage}</Alert></Grid>}
```

frontend/src/pages/Wallet.js
krisutofu: This case might be nearly trivial, but in general it also could mean `provider === undefined`, `provider === ""` etc. or even a combination of them (which would be even worse). I could just search for the documentation of the function, and it says, a promise can be resolved to `null`. I would likely be less uncomfortable with this if the implicit type casting rules of JavaScript wouldn't be broken so much and if it could only mean `provider === null`.

At least if we would like programming style to be consistent, `provider === null` is the better choice because writing conditions with implicit type casts becomes difficult in more complicated cases where the type is not clear.

frontend/src/pages/Wallet.js
krisutofu: Yeah, that's right. From the context, it's clear. I just don't like such abbreviations because it doesn't make sense to abbreviate short names just for the sake of it. It provides no advantage and makes it more difficult to talk about variable. On the other hand, if actual english words are used, it has the advantage that even unfamiliar or non-programming people could know what is talked about. This is at least useful for companies where manager roles work more closely together with developer roles.

frontend/src/pages/Wallet.js
krisutofu: This is a good idea. I think, it would be better to outline it into a function.
```
<OptionalAlert>errorMessage</OptionalAlert>
```

But better would be a variable for that purpose:
```
let errorMessage = ""
//...
errorMessage = <Grid item xs={12}> ... </Grid>
//...
return < ... > errorMessage ... </ ... >
```

frontend/src/pages/Wallet.js
krisutofu: ES6 Classes are just syntactic sugar representing constructor functions. Their only purpose is readability. Maybe, they are not so often used because web developers don't know them much or OOP. My personal advantage is: I can better separate code blocks from longish nested function definitions so that I need less time searching where the code block ends. Don't know, but it helps me to faster understand the code structure.

Also, functions with > 40 lines are often too long. Maybe `EthereumAccountDetails` is at the limit of what can be in a single function. But `Wallet` is already too long for a single function. A `class` would solve this issue.

krisutofu: Better function name should be `EthereumAccountDetails`.

dominic-heil: Please read up on the difference of functional components and class components in React. There is a large difference in how they are written and which functionalities they provide. It is not about some syntactical sugar. Long term React developers know that they can define components as classes, but since the introduction of functional components and hooks rarely ever use them anymore. I do not support rewriting it as a class component. 

Also, as functional components define their own functions to handle different scenarios, I do not support restricting the length of a function to 40 lines as functional components require other functions to implement functionality. Restricting the return statement of a functional component however, or more specifically the number of JSX statements makes sense.

frontend/src/pages/Wallet.js
tuilagio: let us decide it in later phase of the project

frontend/src/pages/Wallet.js
tuilagio: well I tried forEach, for....in, didn't work out, so I stocked with classical for().

frontend/src/pages/Wallet.js
tuilagio: it always returns a list of account, although I couldn't figure out _when_ other accounts are also be returned (I have multiple accounts in my wallets)

dominic-heil (commented):
Getting the ETH from the backend does not seem to work for me. I have run the database setup python script for the eth database collection, uploaded the function to appwrite, set the function ID, and the function is executed when I connected my address (can see that in the appwrite UI), but when I reload the page, the address is not autoloaded. The button says "Connect MetaMask wallet" and after clicking on it the info is retrieved from MetaMask again. Taking a look at the code in my debugger, I found that nothing is returned. After I added an entry to the collection manually for the user with that ID, I finally got a response, but nothing of use was displayed in the frontend, i.e. only the following 

image

Did you have the set-wallet routine working?

tuilagio:
> Getting the ETH from the backend does not seem to work for me. I have run the database setup python script for the eth database collection, uploaded the function to appwrite, set the function ID, and the function is executed when I connected my address (can see that in the appwrite UI), but when I reload the page, the address is not autoloaded. The button says "Connect MetaMask wallet" and after clicking on it the info is retrieved from MetaMask again. Taking a look at the code in my debugger, I found that nothing is returned. After I added an entry to the collection manually for the user with that ID, I finally got a response, but nothing of use was displayed in the frontend, i.e. only the following
> Did you have the set-wallet routine working?
Can you test if the set-wallet works for you? That function is not easy to work with. Remember to set the appwrite URL for that function as the actual IP address (not the public but the local one, i.e 192.168.1.100) otherwise it will not work.

frontend/src/pages/Wallet.js
krisutofu: So, when would this be changed? I think, this should be done now because later we might forget about this and this is important for usability.

frontend/src/pages/Wallet.js
krisutofu, line 37:
```suggestion
                const isEthereumProvider = (provider === null)
		if (!isEthereumProvider) {
```

frontend/src/pages/Wallet.js
krisutofu, line 41: `accountDetails`

tuilagio:
please merge the request, and create an improvement issue. I got enough with this issue.

dominic-heil (approved):
I agree that this should be merged for now. I requested a follow up story, that involves integrating it into the profile, testing and refactoring. Does that sound like a plan @krisutofu?

krisutofu:
Okay, you can merge it on your behalf. This took long already.

krisutofu (commented):
This took too long already, even though I would have liked to see some remaining code changes.

PR #77, opened 2021-11-22
66 implement nft purchase limit

BerinikeTech:


Floinko (approved):
+ Added license
+ Change scope of function from public to internal where possible
+ Meaningful variable names

- New code contains less comments than before. Hard for someone without in-depth knowledge to understand everything. 

In total it looks good to me. Can't think of any more edge cases than the n < 20 one.

blockchain/contracts/AddrArrayLib.sol
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
```
steinkohl, line 1: who is the author, maybe add copyright text?

blockchain/contracts/HelloWorld.sol
```diff
@@ -66,16 +90,20 @@ contract HelloWorld {
...
+            for (
+                uint256 j = 0;
+                j < nftReservations[joinedUsers.getAddressAtIndex(i)];
+                j++
+            ) {
+                nftElement = generateRandomNumber(j) % availableNFTs.length;
+                nftHash = availableNFTs[nftElement];
+                remove(nftElement);
+                nftOwnerships[nftHash].reservedFor = joinedUsers
+                    .getAddressAtIndex(i);
```
steinkohl, line 105: What happens if two contracts are executed at nearly the same time?
If only some or one is left, there is a certain probability.
Raises remove an error if it is not possible?

steinkohl: I misunderstood the operating principle of the solidity techstack - this case is negligible.

steinkohl (approved):
LGTM!

blockchain/contracts/HelloWorld.sol
```diff
@@ -29,16 +37,20 @@ contract HelloWorld {
         nftOwner = payable(0xC77787e364E3420c0609a249F18De47430900f0C);
```
krisutofu, line 37: What is the meaning of this big magic number? Is it possible in Solidity to create a constant for it?

blockchain/contracts/HelloWorld.sol
```diff
@@ -29,16 +37,20 @@ contract HelloWorld {
...
-        uint256 i;
-        // I choose "3" as a value for testing reasons
-        for (i = 0; i < 3; i++) {
+    function createDrop(uint256 _dropTime, uint256 _numberOfNFTS) public {
+        for (uint256 i = 0; i < _numberOfNFTS; i++) {
             mockNFT(_dropTime, i + 1);
         }
+        // We need to let users buy 1 NFT instead of 5% if there are less than 20
+        if (_numberOfNFTS < 20) {
+            fivePercent = 1;
```
krisutofu, line 46: Is there a way to provide a name which is more independent from the hardcoded implementation?
Like
```
purchaseLimit = (_numberOfNFTs * percentLimit) / 100;
nftPurchaseLimit = max(1, purchaseLimit);
```

I don't know however, if such a `max` function exists (for convenience).

blockchain/contracts/HelloWorld.sol
BerinikeTech: It is not a big magic number but an address (to be more specific – Jannis' address). It is hardcoded for testing reasons since the contract is not yet deployed and we also didn't decide yet who the initial owner of the NFTs should be (the person who creates the drop? Or should the person who creates the drop specify this? etc.)

blockchain/contracts/HelloWorld.sol
```diff
@@ -29,16 +37,20 @@ contract HelloWorld {
...
+        } else {
+            fivePercent = (_numberOfNFTS * 5) / 100;
+        }
```
krisutofu, line 49: Is there a way to provide a name which is more independent from the hardcoded implementation?
Like
```
purchaseLimit = (_numberOfNFTs * percentLimit) / 100;
nftPurchaseLimit = max(1, purchaseLimit);
```

I don't know however, if such a `max` function exists (for convenience).

BerinikeTech: Andre decided it should be 5%. If you would like to change that please talk to him during the next meeting :) I don't see a reason why we should force the drop creator to fill out an additional field if it's not necessary.

BerinikeTech: Or were you only talking about the name itself?

krisutofu: Yes, I was talking about the software-engineering part of code like potentially being able to change code easily and the variable name, no new requirement or functionality part.

blockchain/contracts/HelloWorld.sol
krisutofu: Okay, so the meaning is actually: `uint256 jannisAddress = 0xC77787e364E3420c0609a249F18De47430900f0C;`

blockchain/contracts/HelloWorld.sol
BerinikeTech: Will change the names.

blockchain/contracts/HelloWorld.sol
BerinikeTech: No, it is not. The meaning is the current NFT owner is this address and can be payed. It used to be Nick's. As I said, it's only for testing.

blockchain/contracts/HelloWorld.sol
krisutofu: Also for the bottom part, how difficult or easy is it in Solidity to use a more generic variable without hardcoded integer symbols? For example in the error string below.

BerinikeTech: What do you mean by "error string below"?

blockchain/contracts/HelloWorld.sol
```diff
@@ -66,16 +90,20 @@ contract HelloWorld {
...
         uint256 nftElement;
         uint256 nftHash;
-        for (i = 0; i < joinedUsers.size(); i++) {
-            nftElement = generateRandomNumber(i) % availableNFTs.length;
-            nftHash = availableNFTs[nftElement];
-            remove(nftElement);
-            nftOwnerships[nftHash].reservedFor = joinedUsers.getAddressAtIndex(
-                i
-            );
+        for (uint256 i = 0; i < joinedUsers.size(); i++) {
```
krisutofu, line 95: Is there a particular reason to use `uint256` instead of `uint`? Is it faster? Best Practice? The number of people in the world will never fill 64-bit, so 192 bits will always be just zero.

blockchain/contracts/HelloWorld.sol
```diff
@@ -47,17 +59,29 @@ contract HelloWorld {
...
         availableNFTs.push(nftHash);
-        nftsLength = availableNFTs.length;
+        availableNFTsCount = availableNFTs.length;
     }
 
-    function joinDrop() public {
-        // TODO: purchase limit of 5% of all available NFTs
+    function joinDrop(uint256 _numberOfNFTs) public {
         require(
             !joinedUsers.exists(msg.sender),
```
krisutofu, line 67: is `msg` a given variable name? Just asking because of our guidelines and pronounceable variable names.

BerinikeTech: It's given

blockchain/contracts/HelloWorld.sol
BerinikeTech: uint is unint256...

BerinikeTech: And it would be nice if you could leave out unnecessary comments that have absolutely nothing to do with this pr, like the number of observable atoms in our universe :)

blockchain/contracts/HelloWorld.sol
```diff
@@ -47,17 +59,29 @@ contract HelloWorld {
...
         );
-        // We have to make sure that not more users join the drop than we have NFTs available
-        require(joinedUsers.size() <= 3, "You cannot join the drop anymore.");
+        require(
+            reservedNFTsCount != availableNFTs.length,
+            "You cannot join the drop anymore."
+        );
+        require(
+            _numberOfNFTs <= fivePercent,
+            "Sorry, you can't reserve more that 5% of the NFTs."
```
krisutofu, line 76: Does string interpolation or a similar way exist in solidity to turn the `5` into a variable name?

blockchain/contracts/HelloWorld.sol
krisutofu: In line 76

krisutofu:
Well done btw.

blockchain/contracts/HelloWorld.sol
krisutofu: Just saying, about 7/8 of the bitlength likely will never be used, probably even more. I don't know how efficiently large numbers are executed in blockchain nodes. But if there is more efficient data type in terms of computational effort and performance I think it should be prefered.

blockchain/contracts/HelloWorld.sol
krisutofu: It's not forbidden to introduce variables for testing only ;-) . I think, it will also make review more easy.

blockchain/contracts/HelloWorld.sol
BerinikeTech: Unfortunately, I didn't find an easy way to do that (yet) (there are no f strings in Solidity like in Python for example). I think for now, whenever we change the variable which is now called percentageLimit we will have to change this message as well.

Issue #75, opened 2021-11-18
Remove sensitive data from source code repos


krisutofu:
## User story
1. As registered team member
2. I need the specific deployment and operational secrets to be removed from the public repository
3. So that (private) user data is protected.

## Acceptance criteria
* There are no private email addresses, passwords, keys can be found in the repository.
* Sensitive data should not be hardcoded in public automation scripts and source code.
* Sensitive data is instead loaded from private data files, for example CSV / Excel-format.

## Definition of done
* Acceptance criteria are met 
* The code is reviewed by at least two people 
* Every larger component/object/class has a comment/JDoc. Small functions should either be self-explanatory based on the function name or also be commented
* Feature works on the machine of a different team member
* Code is in the main branch
* GitHub CI and checks succeed

PR #79, opened 2021-11-23
75 remove sensitive data

steinkohl:
Removed sensitive data according to #75 and implemented a generic method to create users and add them to teams


backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+
+def add_user_to_team(teams: Teams, user_data: UserData):
+    """
+    Adds user to team, according to the user_data
+    """
+    teams.create_membership(
+        team_id=get_or_create_team_id(teams, user_data.team),
+        email=user_data.email,
+        roles=user_data.roles,
+        url=os.getenv("APPWRITE_ENDPOINT"),
```
dominic-heil, line 120: You cannot use os.getenv here as the user might have used the command line parameters. I only used the command line parameters, which results in `None` for the url parameter and thus failing the operation.

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
-result = users.list()
-print(result)
+    user_data = load_user_data(args.file)
+    for user in user_data:
+        try:
+            add_user(users, user)
+            if len(user.team) > 0:
+                add_user_to_team(teams, user)
+        except AppwriteException as e:
+            print(f"{user.email} - {e}")
```
dominic-heil, line 182: I think, if we get an exception here, then we should continue the for loop and remove the `if len(user.team) > 0` check from here, as it is done below anyways, but this would make the print error messages a little bit more easy to decypher.

dominic-heil (changes requested):
Very well done, works nicely and the code works well! I found one bug, added one suggestion and I think you should also add some sort of Readme. Maybe 1-2 sentences on what the tool does and additionally just copy the print message of the script if you run it with -h?

backend/python_init_script/main.py
steinkohl: Jap, thats a Copy Paste Bug. I will fix that.

backend/python_init_script/main.py
steinkohl: Well, this way already existing users can bei added to Teams as well.

Floinko (approved):
Well done. 
Nothing to add except what Dominic requested.

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+class UserData:
+    """
+    Dataclass for specifying user data
+    """
+
+    name: str
+    email: str
+    team: str
+    roles: [str]
+    init_pw: str
```
krisutofu, line 23: Can you spell out the name as `init_password` or is this a fixed name by convention?

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+        if user_df.shape[1] < 5:
+            user_df = pd.read_csv(file, sep=";", dtype=str)
+    elif ending in ["xlsx", "xls"]:
+        user_df = pd.read_excel(file, dtype=str)
+    else:
+        print("Only csv and xlsx format are supported!")
+
+    for row_id, row in user_df.iterrows():
+        # if cell has no value in xlsx/csv it will result in a nan value
+        # -> replace with empty string
```
krisutofu, line 49: Redundant comments here. From a software-engineering point, these are details cluttered into a bigger function which would better fit into their own functions for better level-of-detail. Functions are very useful for readability so writing small functions is totally fine and even recommended by experienced software engineers.

```
_roles = splitRoles(row)
_team = getTeam(row)
```

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+    """
+    Reads the given file and parses it into UserData objects.
+
+    :param file: path of file containing the user data that shall be loaded
+    :type file: str
+    :return: iterable of the loaded UserData objects
+    :rtype: [UserData]
+    """
+    ending = str(file).lower().split(".")[-1]
+    user_df = pd.DataFrame()
```
krisutofu, line 36: I'm kind of divided how I would rate `user_df` as a name. On the one hand, "dataframe" is a pretty meaningless addition and something like `user_record` would better fit. On the other hand, it's kind of like hungarian notation and it tells you the expected datatype which is fine for Python.

Btw, is it legitimate to write this line in the `else` branch?

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+    if ending == "csv":
+        user_df = pd.read_csv(file, dtype=str)
+        # if there are less than 5 columns -> there might be ';' used as separators
```
krisutofu, line 39: Maybe a poor comment. What about
```
column_count = user_df.shape[1]
if column_count < 5: ...
```

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+        if user_df.shape[1] < 5:
+            user_df = pd.read_csv(file, sep=";", dtype=str)
+    elif ending in ["xlsx", "xls"]:
+        user_df = pd.read_excel(file, dtype=str)
+    else:
+        print("Only csv and xlsx format are supported!")
```
krisutofu, line 45: return statement would fit here

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+    :param teams: teams object, containing the server information
+    :type teams: Teams
+    :param team_name: team name which id shall be returned
+    :type team_name: str
+    :return: ID of the team
+    :rtype: str
+    """
+    result = teams.list()
+    team_ids = [x["$id"] if x["name"] == team_name else None for x in result["teams"]]
+    team_ids = list(filter(None, team_ids))
```
krisutofu, line 103: Also nice could be:
`team_ids = filterTeamIDsByName(team_name, teams.list())`

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
+    If it finds more then one team of the given name, an exception is raised.
+
+    :param teams: teams object, containing the server information
+    :type teams: Teams
+    :param team_name: team name which id shall be returned
+    :type team_name: str
+    :return: ID of the team
+    :rtype: str
+    """
+    result = teams.list()
```
krisutofu, line 101: Has `result` a special meaning? It's not returned below, so I wonder.

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
-# fmt: on
+    client = init_client(api_key=_key, endpoint=_end, project_id=_pid)
+    users = Users(client)
+    teams = Teams(client)
 
-result = users.list()
-print(result)
+    user_data = load_user_data(args.file)
+    for user in user_data:
+        try:
```
krisutofu, line 177: Did you intend this:
```
try:
    add_user()
except AppwriteException as e:
    print ...
if len(user.team) > 0:
    try:
        ...
    except:
        ...
```

When possible, I would avoid nesting structures.

krisutofu (commented):
Good job. It would be good if you could address my comments for a better code quality.

backend/python_init_script/main.py
```diff
@@ -2,41 +2,186 @@
...
-teams.create_membership(team_id=_id, email="admin7@admin1.de", roles=['owner'], url=_endpoint, name="Jannis")
-# fmt: on
+    client = init_client(api_key=_key, endpoint=_end, project_id=_pid)
+    users = Users(client)
+    teams = Teams(client)
 
-result = users.list()
-print(result)
+    user_data = load_user_data(args.file)
+    for user in user_data:
```
krisutofu, line 176: I think, this for-loop would fit well into it's own function `register_user_with_team`

backend/python_init_script/main.py
steinkohl: fixed.

backend/python_init_script/main.py
steinkohl: fixed.
Added get_roles() and more generic _replace_nan() for the rest.
btw, python always uses lower_case_underscore notation for methods, only classes are CamelCase (PEP8)

backend/python_init_script/main.py
steinkohl: here i don't see the improvement, the comment is all about why 5 columns is a messure

backend/python_init_script/main.py
steinkohl: more likely raise an error - i will do so.

backend/python_init_script/main.py
steinkohl: done

backend/python_init_script/main.py
steinkohl: obsolete now

backend/python_init_script/main.py
steinkohl: For this i would have to pass through almost all variables of the main, I don't think that would that much of an improvement.

backend/python_init_script/main.py
krisutofu: Wouldn't you say, it is redundant? Because it tries to explain code. I think your code explains itself well. To make clear that it's about columns you could use `column_count` as name. So after all, variable names just are comments which explain what something does.

backend/python_init_script/main.py
steinkohl: resulting byte code will be identically, but indeed I think thats a bit cleaner

backend/python_init_script/main.py
krisutofu: Okay. It needs `user`, `users`, `teams` and `appwrite_endpoint` but since these variables are "global" you wouldn't need to pass them as parameters to the function.

I only thought about a function name which summarizes what happens. Not mandatory but would be nice.

backend/python_init_script/main.py
krisutofu: I'm sorry, I actually know the the underscore convention in Python but I forgot it :-D .

backend/python_init_script/main.py
```diff
@@ -2,41 +2,215 @@
...
+            x["$id"] if x["name"] == team_name else None for x in _teams.list()["teams"]
+        ]
+        return list(filter(None, _ids))
+
+    team_ids = _filter_team_by_name()
+    # team already exists
+    if len(team_ids) == 1:
+        return team_ids[0]
+    # team not found -> create
+    if len(team_ids) == 0:
```
krisutofu, line 116:
```suggestion
    team_not_found = (len(team_ids) == 0)
    if team_not_found:
```

backend/python_init_script/main.py
```diff
@@ -2,41 +2,215 @@
...
+        _teams: Teams = teams, _team_name: str = team_name
+    ) -> [str]:
+        _ids = [
+            x["$id"] if x["name"] == team_name else None for x in _teams.list()["teams"]
+        ]
+        return list(filter(None, _ids))
+
+    team_ids = _filter_team_by_name()
+    # team already exists
+    if len(team_ids) == 1:
```
krisutofu, line 113:
```suggestion
    team_already_exists = (len(team_ids) == 1)
    if team_already_exists:
```

backend/python_init_script/README.md
```diff
@@ -0,0 +1,16 @@
...
+positional arguments:
+
+    file                   Path of file containing the user data
+
+optional arguments:
+  
+    -h, --help             show this help message and exit
+    --apikey APIKEY        API key to access appwrite backend, you can instead also set the environment variable APPWRITE_API_KEY
+    --endpoint ENDPOINT    URL of the appwrite endpoint, you can instead also set the environment variable APPWRITE_ENDPOINT
+    --projectid PROJECTID  Project ID of the targeted appwrite project, you can instead also set the environment variable APPWRITE_PROJECT_ID
```
krisutofu, line 16: Great

backend/python_init_script/main.py
```diff
@@ -32,33 +32,34 @@ def load_user_data(file: str) -> [UserData]:
     :return: iterable of the loaded UserData objects
     :rtype: [UserData]
     """
+
+    def _replace_nan(val) -> str:
+        # if cell has no value in xlsx/csv it will result in a nan value
+        return "" if str(val) == "nan" else str(val)
+
+    def _get_roles(row):
+        # split roles by ',' if there are more
```
krisutofu, line 41: I'm sorry to comment again. I'm a bit reluctant with the comment because it seems superfluous because the code explains itself already.

backend/python_init_script/main.py
```diff
@@ -32,33 +32,34 @@ def load_user_data(file: str) -> [UserData]:
     :return: iterable of the loaded UserData objects
     :rtype: [UserData]
     """
+
+    def _replace_nan(val) -> str:
+        # if cell has no value in xlsx/csv it will result in a nan value
+        return "" if str(val) == "nan" else str(val)
```
krisutofu, line 38: I also would have liked a function which hides the implementation detail of `nan` like so:
```suggestion
def _get_team(row) -> str:
    return (team := str(row["Team"])) if team != "nan" else ""
```

I think, the code makes clear, that `nan` appears when the string should be empty so that readers don't need to care about 'nan' when reading code in the caller function.

If you like it, you can apply this change.

krisutofu (approved):
Well done! You could take a look at the new comments. I would recommend little more changes but it's your choice.

PR #80, opened 2021-11-24
FAQ page mockup added

NickZecamp:
FAQ


tuilagio (approved):
nice ui

dominic-heil (approved):
Do you also have an .svg of the logos? That would allow us to use it with any resoltion. You can export svg directly from Figma. I already did so for the white logo, but not the rest.

PR #101, opened 2021-11-30
93 multiple drops

BerinikeTech:


blockchain/contracts/NFTtheWorld.sol
```diff
@@ -0,0 +1,180 @@
...
+            _numberOfNFTs <= maxNumberOfNFTsToBuy[_dropHash],
+            "Sorry, you can't reserve more that 5% of the NFTs."
+        );
+        // We have to make sure that not more NFTs get reserved by users than we have NFTs available
+        require(
+            _numberOfNFTs <= availableNFTs[_dropHash].length - reservedNFTsCount[_dropHash],
+            "Sorry, not enough NFTs left for your request"
+        );
+        reservedNFTsCount[_dropHash] += _numberOfNFTs;
+        nftReservations[msg.sender] = _numberOfNFTs;
```
dominic-heil, line 94: The number of reservations of a user is overwritten with each new join. I.e. if I create a drop, join it 5 times, create a new drop, without dropping the old one, join the new one once, then I have only 1 nftreservation. This likely is unexpected behavior.

BerinikeTech: Good catch, will fix :)

steinkohl (approved):
I think i played everything through - LGTM

blockchain/contracts/NFTtheWorld.sol
```diff
@@ -108,20 +117,29 @@ contract NFTtheWorld {
...
         }
     }
 
     // This function lets a user buy her reserved NFTs (one at a time)
     //TODO: Think about a time span during which the reserved NFTs have to be bought
-    function buyNFT(uint256 _price, uint256 _nftHash, uint256 _dropHash) public payable {
+    function buyNFT(
+        uint256 _price,
+        uint256 _nftHash,
+        uint256 _dropHash
```
dominic-heil, line 141: Just curious, why do you use underscore variables?

dominic-heil (approved):
LGTM! Small side note: `generateRandomNumber` does not seem very random. I generated 4 hashes last night in one drop and got two times the same hash. But I think we do not need this method later anyways, so all good :D

steinkohl:
> LGTM! Small side note: `generateRandomNumber` does not seem very random. I generated 4 hashes last night in one drop and got two times the same hash. But I think we do not need this method later anyways, so all good :D

I think we can use the timestamp as well, can't we?

blockchain/contracts/NFTtheWorld.sol
BerinikeTech: It's a common practice to use underscore for the function parameters in solidity :)

PR #106, opened 2021-12-01
88 integrate wallet setting into profile

tuilagio:
Move wallet from pages/ to components/.
Remove "Wallet" button on header.
Add "wallet" area in "Profile" page.
Improve code, add error information.

wallet-in-profile
User with ETH address from backend, Metamask installed
------------------------------
adress-but-no-metamask
User with ETH address from backend, but no Metamask installed
------------------------------
no-address
User without wallet address
------------------------------
no-address-no-metamask
User without wallet address, clicked "Connect" but no Metamask installed
------------------------------


krisutofu:
Nice pictures! Good job, I have a look.

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
+// SPDX-License-Identifier: MIT
+// SDPX-FileCopyrightText: 2021 Que Le <b.le@tu-berlin.de>
+
+
+import appwriteApi from "../api/appwriteApi";
+import { useHistory } from "react-router-dom";
+import { useState } from "react";
+import detectEthereumProvider from "@metamask/detect-provider";
+const { utils } = require( "ethers" );
```
krisutofu, line 9: For myself: What benefit does `require` have over import? Is there one?

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	
+	return <TableRow>
+		<TableCell style={{ color: "white", borderBottom: "none" }}>
+			{data.pubAddress}
+		</TableCell>
+		<TableCell style={{ color: "white", borderBottom: "none" }}>
+			{balance}
+		</TableCell>
+	</TableRow>;
+}
```
krisutofu, line 41: Nice function!
Only one usability thing: users might not understand or like to read "NaN". Maybe you could use "--" or "unknown" as fallback text?

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	useEffect(() => {
+		fetchAccountsDetails(publicAddresses);
+	}, []);
+
+	return <TableContainer style={{ color: "white" }}>
+		<Table>
+			<TableBody>
+				<TableRow>
+					<TableCell style={{ color: "white", borderBottom: "none" }}>
+						<Typography variant="subtitle1" gutterBottom component="div">
```
krisutofu, line 88: I think, this should be `ParagraphTypography` because of design. It would be quite low effort to change it in this PR compared to someone changing this later in an upcoming PR (where git-diffs get larger).

Therefore, I'd like to have `ParagraphTypography` instead of `Typography`.

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+
+
+import appwriteApi from "../api/appwriteApi";
+import { useHistory } from "react-router-dom";
+import { useState } from "react";
+import detectEthereumProvider from "@metamask/detect-provider";
+const { utils } = require( "ethers" );
+import { delayMsec } from "../utils/utils";
+
+import Typography from "@mui/material/Typography";
```
krisutofu, line 12: It would be better to import our own ParagraphTypography here...

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	</TableContainer>;
+}
+
+/**
+ * Component to display and add Eth address from Wallet. Currently only Metamask is supported.
+ * @param user user object of the currently logged in user/admin
+ * @returns {JSX.Element}
+ */
+export default function Wallet({ user }) {
+	const [loadedAddresses, setLoadedAddresses] = useState(false);
```
krisutofu, line 112: Variable's name sounds like an array to me instead of a boolean. I would prefer, if variable names of a boolean variable would sound like a predicate or question. A common convention of naming booleans is using `is` or `has` as name prefix or `Flag` as suffix.

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	const routeChange = (path) =>{
+		history.push(path);
+	};
+
+	const getEthAddressesFromServer = () => {
+		if (!loadedAddresses) {
+			appwriteApi.getOwnEthAddress(user.$id)
+				.then(result => {
+					// Set the address(es)
+					let currentPubAddr = publicAddresses;
```
krisutofu, line 127: Spelling out this variable name would be at least as fine. Don't be afraid of long names. You can use autocompletion.

Spelling out `pubAddr` in line 157 would be even better.

frontend/src/components/Wallet.js
dominic-heil: https://stackoverflow.com/a/46677972/7838440 answers your question pretty well. We should not use require, but import instead. You can run `npm i ethers` to add it to the package.json and your downloaded node_modules

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+					for (let i = 0; i < result.sum; i++) {
+						currentPubAddr.push(result.documents[i].walletAddress);
+					}
+					setPublicAddresses(currentPubAddr);
+					setLoadedAddresses(true);
+				})
+				.catch((error) => {
+					console.error(error);
```
krisutofu, line 135: Should this error message be hidden from users?

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	useEffect(() => {
+		getEthAddressesFromServer();
+	}, []);
+
+	if (!user){
+		routeChange("/");
+		return <></>;
+	}
+
+	/* Handle button connect to MetaMask */
```
krisutofu, line 149: This comment actually is a function name at least by commonly taught software engineering.
Same for line 158.
Comments are not bad but their purpose is different from this. If you'd like to explain what your code snippet does (which is very legitimate) then function names are a better choice.

Single-statement functions are perfectly fine, when there is a need to comment that statement.

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+		if (ethereum) {
+			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
+			const account = accounts[0];
+			pubAddr = account;
+			// Set address on server side
+			appwriteApi.setEthAddress(pubAddr);
+
+			let currentPubAddr = publicAddresses;
+			currentPubAddr.push(pubAddr);
+			setPublicAddresses(currentPubAddr);
```
dominic-heil, line 163: This does not have the expected behavior as you cannot edit a state this way. In fact, it does not work this way for me. Instead do this:
```suggestion
			setPublicAddresses(prevAddr => [...prevAddr, pubAddr]);
```

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+			{balance}
+		</TableCell>
+	</TableRow>;
+}
+
+function EthereumAccountDetails ({ publicAddresses, setErrorNoMetamaskMessage }) {
+	const [detailData, setDetailData] = useState([]);
+
+	async function fetchAccountsDetails(publicAddresses) {
+		// TODO: publicAddresses is still not loaded.
```
dominic-heil, line 47: Here, check if publicAddresses is loaded and if not just return or throw an exception. Below, in the useEffect, add `publicAddresses` in the square brackets. The function in useEffect is called every time any of the states in the square brackets changes. Thus, whenever publicAddresses, this function will be called. Contact me if you would like some more clarifications.

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	</TableRow>;
+}
+
+function EthereumAccountDetails ({ publicAddresses, setErrorNoMetamaskMessage }) {
+	const [detailData, setDetailData] = useState([]);
+
+	async function fetchAccountsDetails(publicAddresses) {
+		// TODO: publicAddresses is still not loaded. 
+		// Solve temporarily by waiting for 0.5 sec.
+		await delayMsec(500);
```
dominic-heil, line 49: Please removed this. You rarely ever want to lazily wait for something anywhere.

dominic-heil (changes requested):
Functionality wise works fine with the small proposed changes. Good job! Once the things I mentioned are changed you will get my approval =)

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+			appwriteApi.setEthAddress(pubAddr);
+
+			let currentPubAddr = publicAddresses;
+			currentPubAddr.push(pubAddr);
+			setPublicAddresses(currentPubAddr);
+		} else {
+			setErrorNoMetamaskMessage("Please install MetaMask!");
+			console.log("Please install MetaMask!");
+			return;
+		}
```
krisutofu, line 168: Suggestion:
```suggestion
			if (ethereum === null) {
				return setErrorNoMetaMaskMessage("Please install MetaMask!");
			}
			const accounts = await ethereum.request({ method: "eth_requestAccounts" });
			const newPublicAddress = accounts[0];
			setWalletAddressOnServer(newPublicAdress);
			setPublicAddresses(old => {...old, newPublicAddress});
```

frontend/src/components/Wallet.js
```diff
@@ -0,0 +1,184 @@
...
+	};
+
+	return <Grid item style={{ width: "100%" }}>
+		{publicAddresses.length==0
```
krisutofu, line 172: Switching if-branch with else-branch would be more fault-tolerant (imagine publicAddresses would be `undefined` in a weird scenario):
`publicAddresses.length !== 0`

frontend/src/utils/utils.js
```diff
@@ -14,4 +14,15 @@ export const isValidEmail = (email) => {
...
+ * @param ms miliseconds to sleep
+ * Source: https://stackoverflow.com/a/47480429
+ * Usage:
+ * import delayMsec from "../utils";
+ * const yourFunction = async () => {
+ *     await delayMsec(5000);
+ *     console.log("Waited 5s");
+ *};
+ */
+export const delayMsec = ms => new Promise(res => setTimeout(res, ms));
```
krisutofu, line 28: Great that you put this into `utils.js`. I wonder, whether there would have been an existing function that you could have used instead.

krisutofu (commented):
There are some things that I would like to see improved but I found no particular bugs or big problems.

frontend/src/components/Wallet.js
krisutofu: Coincidentally, this is included in my code suggestion below :-) .

dominic-heil: Have you tested yours? You create an object, even though we need a list. I tested mine.

frontend/src/components/Wallet.js
krisutofu: You mean busily waiting, right?

frontend/src/components/Wallet.js
krisutofu: I edited my example, I also saw the mistake :-) .

frontend/src/components/Wallet.js
dominic-heil: Busy waiting would be something like 
```
while (!condition) {
 // do nothing
}
// do stuff that requires condition to be true
```

I thought what I mean is called lazy waiting, but I must have been wrong. Anyways, sleeping at any point in the code (or buzy waiting) should not be done anywhere. Instead, react to something like events that are fired when a state changes, as in the case of a state chaning which results in useEffect being triggered.

frontend/src/components/Wallet.js
tuilagio: I would keep that. My eyes after hours got tired distinguishing `currentPublicAddress` and `publicAddresses`.

frontend/src/components/Wallet.js
tuilagio: this could have reduced around 3 hours working for me. How could I oversee this ^^

frontend/src/components/Wallet.js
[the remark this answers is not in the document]
tuilagio: no I will keep this one. Too tired now

frontend/src/components/Wallet.js
tuilagio: well it can't be null. Even if it was, `publicAddresses.length` will not save you.

dominic-heil (approved):
LGTM! Thank you for the changes!

frontend/src/components/Wallet.js
tuilagio: fixed

frontend/src/components/Wallet.js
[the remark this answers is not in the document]
krisutofu: Just know that doing so creates unnecessary changes later.

frontend/src/components/Wallet.js
krisutofu: You could have named it `currentWalletAddress`, easy to distinguish.

frontend/src/components/Wallet.js
krisutofu: Okay, unlikely, but if `publicAddress` would be undefined or null, the condition would evaluate to false in any case so that the error handling suits more to the else branch.

krisutofu (commented):
Some things that I find good, some that I find improvable. I'm too tired to approve it now.

krisutofu (approved):
Okay, you worked hard, here you get your approval. It's not the final design ;-) .

PR #118, opened 2021-12-05
Add the Welcome Banner to Frontend

krisutofu:
This is my first visual outcome, a simple incomplete version of the banner.

I had a lot of trouble implementing the design despite using the data from Figma.

Problem:
- The text thickness is not right.
- The profile button does not appear when logged in as Admin (also user?).
- The background image repeats instead of stretches on wide screen (I don't have the original background image).
- Border does not have rounded corners which I set for the container.
- Design failure.

You can give me feedback or criticism towards code or style. Keep in mind, this code is only provisoric. I would remove Code Comments from a non-draft PR.

You can see the result in the FAQ page. It at least similes the original design slightly.
Screenshot from 2021-12-05 03-07-25

(I took extremely long to get working on this. I was thinking too much. At least JavaScript basically compiles at the first try, except for some initial import problems, so it's less frustrating.)


frontend/src/assets/jss/colorPalette.js
```diff
@@ -6,3 +6,4 @@ export const textColor = "#FFFFFF";
 export const secondaryTextColor = "#D2D2D2";
 export const activeTextColor = "#009C19";
 export const backgroundColor = "#1C1C1C";
+export const signupButtonColor = "#009C19";
```
krisutofu, line 9: This would also be useful for later refactoring, for example replacing hard-coded values from the headers.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+			}</Grid>
+		</div>
+	);
+}
+/*
+export default function Banner({ children, title, user }) {
+	const [isAdmin, setAdminStatus] = useState(false);
+	const isLoggedIn = (user != null);
+
+	const updateAdminStatus = async () => {
```
krisutofu, line 54: Please ignore "async" here, didn't bother to delete it. Would require `useAsyncEffect`.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+
+function SignupButton(style) {
+	return (
+		<RoundedEdgesButton style={{ backgroundColor: signupButtonColor, ...style }} component={Link} to="/signup">
+			Sign Up
+		</RoundedEdgesButton>
+	);
+}
+
+function ProfileButton({ style }) {
```
krisutofu, line 106: Ah, I just see why the button is not working. Will fix it tomorrow.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+// according to mockup, the Banner's Button is slightly larger than the Header's
+const bannerButtonStyle = {
+	height: 54,
+	width: 192,
+	left: 624,
+	top: 303,
+	borderRadius: 27,
+};
+
+function BannerButton({ isLoggedIn }) {
```
krisutofu, line 91: Forgot `style` parameter

frontend/src/assets/jss/colorPalette.js
dominic-heil: Two lines above we define the same color which we use in the header buttons.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+import RoundedEdgesButton from "./RoundedEdgesButton";
+import { Grid } from "@mui/material"
+
+import { textColor, signupButtonColor } from "../assets/jss/colorPalette";
+import bannerImage from "../assets/img/banner.png"
+
+const backgroundImageStyle = {
+	backgroundImage: `url(${bannerImage})`,
+	backgroundPosition: "center",
+};
```
dominic-heil, line 18:
```suggestion
const backgroundImageStyle = {
	backgroundImage: `url(${bannerImage})`,
	backgroundPosition: "center",
	backgroundSize: "cover"
};
```
to have the image always cover the whole div and not repeat itself. Also works great for responsive designs.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+		padding: 40,
+		textAlign: "center",
+	};
+
+	return (
+		<div style={backgroundImageStyle}>
+			<Grid container direction="column" alignItems="center" justifyContent="center" columnSpacing={28} style={bannerContainerStyle}>{
+				[bannerTitle, bannerSubTitle, bannerButton].map((item,index) => (
+					<Grid item key={index}>{item}</Grid>
+				))
```
dominic-heil, line 44: This does not improve readability at all. Just put in the three components here.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+		borderRadius: 15,
+		padding: 40,
+		textAlign: "center",
+	};
+
+	return (
+		<div style={backgroundImageStyle}>
+			<Grid container direction="column" alignItems="center" justifyContent="center" columnSpacing={28} style={bannerContainerStyle}>{
+				[bannerTitle, bannerSubTitle, bannerButton].map((item,index) => (
+					<Grid item key={index}>{item}</Grid>
```
dominic-heil, line 43: I don't think this should be a grid. Just put in the three components here and add some paddings/margins above/below them. This grid also breaks the responsive design.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+	const bannerButton = <BannerButton isLoggedIn={(user != null)} style={{ fontSize: 17, }}/>;
+	const bannerContainerStyle = {
+		height: 305,
+		width: 1100,
+		left: 164,
+		top: 96,
+		borderRadius: 15,
+		padding: 40,
+		textAlign: "center",
+	};
```
dominic-heil, line 37: Do not hardcode any width. This destoys responsive designs. Instead use something like `100%` or `100vw` or other ratios.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function BannerSubTitle({ style, children }) {
+	return <ParagraphTypography style={{ color: textColor, fontSize: 22, ...style }}>{ children }</ParagraphTypography>
+}
+
+// according to mockup, the Banner's Button is slightly larger than the Header's
+const bannerButtonStyle = {
+	height: 54,
+	width: 192,
```
dominic-heil, line 85: Width and height must be px values, e.g. `height: "54px"`

frontend/src/components/Banner.js
dominic-heil: Also, personaly I would just put these style elements into the component directly. But thats mostly up to you.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,112 @@
...
+	return <ProfileButton style={bannerButtonStyle}/>;
+}
+
+function SignupButton(style) {
+	return (
+		<RoundedEdgesButton style={{ backgroundColor: signupButtonColor, ...style }} component={Link} to="/signup">
+			Sign Up
+		</RoundedEdgesButton>
+	);
+}
```
dominic-heil, line 104: Font size of this button is not correct. Also you forgot the curly brackets around style.

dominic-heil (commented):
Just a few preliminary comments. Still needs improvements. Please try to make it so that the banner also looks ok on devides with width 320-600px.

frontend/src/assets/jss/colorPalette.js
krisutofu: Right, I didn't notice. I better put `const signupButtonColor = activeTextColor` into the `SignupButton` component, right?

frontend/src/components/Banner.js
krisutofu: I wish there would be a way to avoid this code duplication. For better maintenance, y'know?

frontend/src/components/Banner.js
krisutofu: I liked to omit the curly braces but was I inconsistent in doing so and it caused the profile button to "not appear".

I am aware, JSX fnotation would be different. But since they are private functions, I thought, it's not a problem, no?

krisutofu:
Screenshot from 2021-12-05 21-35-08
)

frontend/src/components/Banner.js
krisutofu: Done

frontend/src/components/Banner.js
krisutofu: Done

frontend/src/components/Banner.js
krisutofu: Done (?)

frontend/src/components/Banner.js
krisutofu: I removed it, done

frontend/src/components/Banner.js
krisutofu: I removed it.

frontend/src/components/Banner.js
krisutofu: Thank you a lot, I use this.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,108 @@
...
+
+const newBannerStyle = (style) => ({ color: textColor, ...style, });
+
+function BannerTitle({ style, children }) {
+	const smallStyle = newBannerStyle({ fontSize: 30, fontWeight: 550, bottomMargin: "5px", ...style });
+	const mediumStyle = newBannerStyle({ fontSize: 37, fontWeight: 700, ...style });
+
+	return <HeaderTypography sx={smallStyle} md={mediumStyle}>{ children }</HeaderTypography>;
```
krisutofu, line 66: Seems like `xs` and `md` are not working for this component?

frontend/.env
```diff
@@ -6,6 +6,7 @@ ESLINT_NO_DEV_ERRORS=true
 
 REACT_APP_ENDPOINT="http://localhost:80/v1"
 REACT_APP_PROJECT=""
+REACT_APP_DOMAIN="http://localhost:3000"
```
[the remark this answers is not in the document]
krisutofu: `domainName` in `frontend/src/utils/config.js` is not an environment variable. The environment variable would be undefined.
I think you shouldn't have removed the definition of `REACT_APP_DOMAIN` because it makes the build process less transparent, less clear for new people. The other environment variables are defined in `./frontend/.env` and therefore `REACT_APP_DOMAIN_NAME` should be there as well.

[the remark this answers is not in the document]
dominic-heil: I agree. Maybe at some point we should also create a .env.defaults or so.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,108 @@
...
+		height: "54px",
+		width: "192px",
+		fontWeight: 700,
+		...style,
+	});
+
+	if (!isLoggedIn) {
+		return <SignupButton style={buttonStyle}/>;
+	}
+	return <ProfileButton style={buttonStyle}/>;
```
dominic-heil, line 88: I think something like the following would look nicer. But can also stay the way it is.
```JavaScript
let BannerButtonComponent = isLoggedIn ? ProfileButton : SignupButton;
return <BannerButtonComponent style={buttonStyle}/>;
```

frontend/src/components/Banner.js
dominic-heil: `sx` is the custom style property of MUI and `md` is not defined for HeaderTypography. React does not support `md` out of the box and for us only works for MUI components that know how to handle it. You can use something like [react-responsive](https://www.npmjs.com/package/react-responsive) to find out if the window width is above a certain threshold and based on that change the style.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function SignupButton({ style }) {
+	const backgroundColor = activeTextColor;
+	return (
+		<RoundedEdgesButton style={{ backgroundColor, ...style }} component={Link} to="/signup">
+			Sign Up
+		</RoundedEdgesButton>
```
dominic-heil, line 96:
```
style={{ backgroundColor: activeTextColor, ...style }}
```
is easier to read imo.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,108 @@
...
+import bannerImage from "../assets/img/banner.png";
+
+/**
+ * Displays a welcoming background image with a button.
+ * @param user user object of the currently logged in account
+ * @returns {JSX.Element}
+ */
+export default function Banner({ children, title, user }) {
+	const bannerTitle = <BannerTitle>{ title }</BannerTitle>;
+	const subTitle = <BannerSubTitle>{ children }</BannerSubTitle>;
```
dominic-heil, line 20: I would rename `children`to `subtitle` as `children` is usually used to wrap any components but here we only expect text.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,108 @@
...
+	const button = <BannerButton isLoggedIn={(user != null)} style={{ fontSize: "17px", }}/>;
```
dominic-heil, line 21: I see no advantage of defining these varialbes if we could just include them in the component. Also, you could just include the fontStyle style in the button component itself as we only use that component once anyways.

dominic-heil (commented):
Component looks good. Good job! Apart from the different sizes of the baned based on the banner width that you did not working, I have no functional complains. Please try to make it so that the banned also looks good on devices with width at least 320px.

krisutofu:
Thank you Dominic for all your comments. I close this early draft. I already implemented most of your proposed changes and will commit it soon.

Issue #113, opened 2021-12-02
Secure NFT-drops to be created only by authorized addresses


krisutofu:
## User story
1. As a market/application owner
2. I need security measures in the blockchain
3. So that users are not scammed and have a better experience

## Acceptance criteria
* It can be demonstrated that non-elligable addresses can not create NFT-drops.
* It can be demonstrated that elligible addresses can create NFT-drops.
* There is a way to add new authorized addresses.

## Definition of done
* The code is reviewed by at least two people 
* Every larger component/object/class has a comment/JDoc. Small functions should either be self-explanatory based on the function name or also be commented
* Feature works on the machine of a different team member
* Code is in the main branch

Issue #110, opened 2021-12-02
Add price per NFT to createDrop


BerinikeTech:
## User story
1. As a drop creator 
2. I want to specify a price per NFT
3. So that I add monetary value to my artwork

## Acceptance criteria
* During drop creation the creator can specify a price
* All other functions using price are adapted 
* A user who wants to buy an NFT has to pay the specified price

## Definition of done
* The code is reviewed by at least two people 
* Every larger component/object/class has a comment/JDoc. Small functions should either be self-explanatory based on the function name or also be commented
* Feature works on the machine of a different team member
* Code is in the main branch
* GitHub CI and checks succeed

PR #119, opened 2021-12-06
Price per NFT & User Access

JaPilgrim:
Testing of Features & functionalities of #110 & #113 succeeds
I created & verified the source code, just commented out the "constructor". If now deployed with the "constructor", it gives the following messages and prohibits a re-verification for the contract WITH the constructor. 
image
Therefore I assume it to be verifiable.

### DIFF
--- a/blockchain/contracts/NFTtheWorld.sol
+++ b/blockchain/contracts/NFTtheWorld.sol
@@ -3,8 +3,13 @@
 
 pragma solidity ^0.8.0;
 
+import "./AddrArrayLib.sol";
+
 contract NFTtheWorld {
+    using AddrArrayLib for AddrArrayLib.Addresses;
     address payable nftOwner;
+
+    AddrArrayLib.Addresses adminAddresses;
 
     uint256 percentageLimit;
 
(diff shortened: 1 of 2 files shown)


blockchain/contracts/NFTtheWorld.sol
```diff
@@ -3,23 +3,31 @@
...
-        address owner;
+        address payable owner;
         uint256 nftId;
         uint256 dropId;
         uint256 dropTime;
+        uint256 weiPrice;
         address reservedFor;
     }
 
+    mapping(address => bool) public adminRights;
```
dominic-heil, line 29: Does this need to be public? I don't think it destroys anything, just curious.

blockchain/contracts/NFTtheWorld.sol
```diff
@@ -3,23 +3,31 @@
 
 pragma solidity ^0.8.0;
 
+//import "./AddrArrayLib.sol";
+
+
 contract NFTtheWorld {
     address payable nftOwner;
 
+    address payable[] adminAddresses;
```
dominic-heil, line 12: This variable was created but is never used?

dominic-heil (commented):
Code looks fine. I tried to run the adminRights function on the blockchain (https://kovan.etherscan.io/address/0x2728c74325b2c3280cabc662dc3b05c906b8ae17#readContract) with my address and am not an admin as expected. I also tried to run the adminRights function with the contract creator address (0xe016160abedf3537bc3d7ba946a4dba3ee49ceed) but that address also was no admin. So I cannot confirm if the code works as intended.

blockchain/contracts/NFTtheWorld.sol
JaPilgrim: Hmm no not in general, but for right now & testing purposes it suits good I think

blockchain/contracts/NFTtheWorld.sol
```diff
@@ -172,6 +193,23 @@ contract NFTtheWorld {
...
+
+    function addToAdmins(address payable _addressToAdd) public onlyByAdmins() {
+        require(!adminRights[msg.sender]=true, "Address already has Admin rights.");
+        adminRights[_addressToAdd]=true;      
+    }
+
+    function removeFromAdmins(address payable _addressToRemove) public onlyByAdmins() {
+        require(!adminRights[msg.sender]=false, "Address already has no Admin rights.");
+        adminRights[_addressToAdd]=false;  
+    }
```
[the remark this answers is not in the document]
JaPilgrim: Good idea in general. I added it, but for testing I commented it out.

JaPilgrim:
PR closed & code integrated into branch #120

Issue #68, opened 2021-11-18
Implement mockups designs of main/landing page 


dominic-heil:
## User story
1. As a NFT interested website user
2. I want the main landing page to look promissing
3. So that I am convinced by the product and NFTs quickly 

## Acceptance criteria
* The general design of the main landing page looks like in the mockups
* Texts and images may be mocked or left empty until we have real data. 

## Definition of done
* The code is reviewed by at least two people 
* Every larger component/object/class has a comment/JDoc. Small functions should either be self-explanatory based on the function name or also be commented
* Feature works on the machine of a different team member
* Code is in the main branch

PR #121, opened 2021-12-06
#68 Add Landing Page with Mockup Design

krisutofu:
Now with simplistic landing page.

I use an orange placeholder box for the NFT-drop lists.
I'm not entirely sure, how André will want this to be. Let's just implement it and then ask.

This is how it looks:
Screenshot from 2021-12-06 23-32-23
Screenshot from 2021-12-06 23-33-39


frontend/src/components/Banner.js
```diff
@@ -0,0 +1,107 @@
...
+				<BannerItem style={{ }}>{ button }</BannerItem>
+			</div>
+		</div>
+	);
+}
+
+const bannerStyle = ({ color: textColor });
+
+function BannerTitle({ style, children }) {
+	//const smallStyle = { ...bannerStyle, fontSize: 30, fontWeight: 550, bottomMargin: "5px", ...style };
```
krisutofu, line 67: I didn't know here how to apply it for the small case. `xs` and `md` didn't work.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// according to mockup, the Banner's Button is slightly larger than the Header's
+function BannerButton({ isLoggedIn, style }) {
+	const buttonStyle = { height: "54px", width: "192px", fontWeight: 700, ...style };
+
+	if (!isLoggedIn) {
+		return <SignupButton style={buttonStyle}/>;
+	}
+	return <ProfileButton style={buttonStyle}/>;
```
dominic-heil, line 87: I think something like the following would look nicer. But can also stay the way it is.
```JavaScript
let BannerButtonComponent = isLoggedIn ? ProfileButton : SignupButton;
return <BannerButtonComponent style={buttonStyle}/>;
```

frontend/src/components/Banner.js
dominic-heil: `sx` is the custom style property of MUI and `md` is not defined for HeaderTypography. React does not support `md` out of the box and for us only works for MUI components that know how to handle it. You can use something like [react-responsive](https://www.npmjs.com/package/react-responsive) to find out if the window width is above a certain threshold and based on that change the style.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,107 @@
...
+
+function BannerTitle({ style, children }) {
+	//const smallStyle = { ...bannerStyle, fontSize: 30, fontWeight: 550, bottomMargin: "5px", ...style };
+	const mediumStyle = { ...bannerStyle, fontSize: 37, fontWeight: 700, ...style };
+
+	return <HeaderTypography style={mediumStyle}>{ children }</HeaderTypography>;
+}
+
+function BannerSubTitle({ style, children }) {
+	//const smallStyle = { ...bannerStyle, fontSize: 19, fontWeight: 400, ...style };
```
dominic-heil, line 74: dead code

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Displays a welcoming background image with a button.
+ * @param user user object of the currently logged in account
+ * @returns {JSX.Element}
+ */
+export function Banner({ children, title, user }) {
+	const bannerTitle = <BannerTitle>{ title }</BannerTitle>;
+	const subTitle = <BannerSubTitle>{ children }</BannerSubTitle>;
+	const button = <BannerButton isLoggedIn={(user != null)} style={{ fontSize: "17px", }}/>;
```
dominic-heil, line 30: I see no advantage of defining these varialbes if we could just include them in the component. Also, you could just include the fontStyle style in the button component itself as we only use that component once anyways.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * Displays a welcoming background image with a button.
+ * @param user user object of the currently logged in account
+ * @returns {JSX.Element}
+ */
+export function Banner({ children, title, user }) {
+	const bannerTitle = <BannerTitle>{ title }</BannerTitle>;
+	const subTitle = <BannerSubTitle>{ children }</BannerSubTitle>;
```
dominic-heil, line 29: I would rename `children`to `subtitle` as `children` is usually used to wrap any components but here we only expect text.

dominic-heil (commented):
Component looks good. Good job! Apart from the different sizes of the baned based on the banner width that you did not working, I have no functional complains. Please try to make it so that the banned also looks good on devices with width at least 320px and text does not go under e.g. buttons but instead increases the height of the component. You might want to work with `minHeight` instead of `height`.

Also, next time just continue working on the other PR/draft or create the second draft/PR to merge into the first to keep the dif smaller and easier to check.

frontend/src/components/Banner.js
krisutofu: Yes, both is legitimate. In my case, I aimed at easy maintainability, that means easier to change because who knows the future ;-) ? Also main reasons basically are lower indentation width and more levels of detail. But honestly I don't know how useful it really is in this case.

krisutofu:
The banner styles for small screens (below 900px) has been added and looks good for me at least.

I think, I'm done. If you approve this today and we merge it, then we could show it tomorrow morning.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,110 @@
...
+		<div style={containerStyle}>
+			<div style={backgroundImageStyle}>
+				<div style={{ ...bannerItemStyle, marginBottom: "11px" }}>{ bannerTitle }</div>
+				<div style={{ ...bannerItemStyle, height: "66px", marginBottom: "32px" }}>{ bannerSubtitle }</div>
+				<div style={{ ...bannerItemStyle, }}>{ button }</div>
+			</div>
+		</div>
+	);
+
+	const largeSizeThreshold = 900;
```
dominic-heil, line 45: I would change this to 600, as between 600-900 nothing breaks and it still looks good.

frontend/src/components/Banner.js
```diff
@@ -0,0 +1,110 @@
...
+	const bannerSubtitle = <BannerSubTitle isLarge={isLarge}>{ subtitle }</BannerSubTitle>;
+	const button = <BannerButton isLoggedIn={(user != null)} isLarge={isLarge}/>;
+
+	const containerStyle = ({ minHeight: "305px", });
+	const backgroundImageStyle = ({
+		padding: "44px",
+		alignItems: "center",
+		justifyContent: "center",
+		textAlign: "center",
+
```
dominic-heil, line 58: maybe remove empty line and maybe make it a one liner?

dominic-heil (approved):
LGTM! No real complains. Good job! :D

frontend/src/components/Banner.js
krisutofu: Okay, I found the sweet spot which is 700px. When I tested it in Firefox with 600px, it will reach temporarily 3 subtitle lines. 700px closely avoids that.

frontend/src/components/Banner.js
krisutofu: Haha, this is now a long line. At the beginning I wasn't sure whether I should do this but this is consistent :-D .

All those properties would otherwise take a lot of lines and are usually not-so-interesting details. By writing them on one line it's like collapsed.

tuilagio (approved):

PR #122, opened 2021-12-07
fixed init script

steinkohl:
fixed: print env variables before exiting with error message
edited: removed _ID from appwrite project env variable as suggested from @dominic-heil


dominic-heil (approved):
LGTM

PR #124, opened 2021-12-08
112 implement proper design for announcements

tuilagio:
This PR cleans up the Announcement code and fix the UI.
To add and display picture for announcement is another follow up task needed.
Failed to implement `read more` link in each reduced text.
------------------------------
Not admin

a1

a2

------------------------------
Admin view

a1_admin

a2_admin


dominic-heil (changes requested):
Just looking at the screenshots you provided and by running it locally on machine, your implemented design does not look close to the mockups.

Ignoring that you have no images (which for now is fine I think):
- font family and weight of titles does not match
- date format is different
- previews are longer than two lines. There is a way to tell  `<p>` to never go more than e.g. 2 lines using something like the following css style attributes:
  ```
  overflow: hidden;
  display: -webkit-box;
  -webkit-line-clamp: 5;
  -webkit-box-orient: vertical;
  ```
- the "read more" is italic and blue instead of non-italic and white.
- the font-family and -color of the previews texts is different
- the divider between the announcements is differently colored
- the space between two announcements is higher than in the mockups (99px vs 129px)

krisutofu:
> Just looking at the screenshots you provided and by running it locally on machine, your implemented design does not look close to the mockups.
> 
> Ignoring that you have no images (which for now is fine I think) a for now non exhaustive list:
> 
>     * font family and weight of titles does not match
> 
>     * date format is different
> 
>     * previews are longer than two lines. There is a way to tell  `<p>` to never go more than e.g. 2 lines using something like the following css style attributes:
>       ```
>       overflow: hidden;
>       display: -webkit-box;
>       -webkit-line-clamp: 5;
>       -webkit-box-orient: vertical;
>       ```
> 
>     * the "read more" is italic and blue instead of non-italic and white.
> 
>     * the font-family and -color of the previews texts is different
> 
>     * the divider between the announcements is differently colored
> 
>     * the space between two announcements is higher than in the mockups (99px vs 129px)

I agree on all of this. Yet it's a draft but actually would have been nice if this could have been drafted earlier. I am not willing to make through night for this PR.

krisutofu:
As for the time format: you chose an Brittisch date format. Could you use German date for Germany, Brittisch for rest of Europe and US for America?

krisutofu:
Your sidebar still is not displayed properly. The "Announcements" heading is more indentend than the single announcements and the whole sidebar is padded (but should not be padded).
Here is the image:
Screenshot from 2021-12-08 20-44-33

krisutofu (changes requested):
Looks better now and closer to the mockups. But the acceptance criteria are not fully satisfied. There need to be images. You can use placeholder boxes with the images that you exported from Figma. But also the Announcement Editor page should display image boxes for announcements.

You don't need image uploading functionality or the ability to change images. These images can be hardcoded (every n-th announcement gets `announcement_${n}.png` or randomly selected and displayed). Uploading will be done later when we can upload images for creating NFTs.

krisutofu:
My date still displays a weird year. Is this related to the database collection?

krisutofu:
Would it also make sense to update the date, when changes of an announcement are submitted? I think, this would be a very easy change.

Is the submit button supposed to keep the text-editor open like the current behaviour? If yes, I would rename the "submit" button to "apply" and the "cancel" button to "finish" or "done".

krisutofu:
I also would like to have consistency of the Announcement editor page with the rest of the App for better Admin experience. This means using `ParagraphTypography` or `HeaderTypography` instead of `Typography` for every text and using the proper black or green button design.
Use a red colour for the "Delete" button text+border and maybe white or green color for the "Edit" button text+border.

dominic-heil (changes requested):
The announcements still do not look close to the mockups:
- the space between title and date, and date and description is off
- the font color of the description is still white and not `rgba(255, 255, 255, 0.81);` as in the mockups
- date format is different (and still bugged for me:
    image
- "Read more" is not displayed. Maybe the line-clamp approach does not work, but there is a JS solution for this: https://stackoverflow.com/a/41549042/7838440
- date color is also off (you have 0.5 opacity, mockups has 0.4)
- the divider color is off (mockups use `rgba(255, 255, 255, 0.1)`)
- the space between the "Announcements" heading and the banner is off.

Please look at the mockups in Figma again.

dominic-heil:
Accepting this now to fix the main. Please create a new PR for fixes.

PR #125, opened 2021-12-08
89 profile page mockup

NickZecamp:
Profile
NFT list (1)
NFT list - mobile
Profile - Mobile
profile page and user NFT list page


krisutofu (approved):
Hi Nick, would be nice if you could enter some reviewers for the next design stories.

Profile is well-done.

If the profile would be viewed by other logged in people, what would be different from the current images? I think, the text fields would be fixed, non-editable, right?

steinkohl (approved):
LGTM

PR #126, opened 2021-12-09
integration test workflow

Floinko:


dominic-heil (commented):
Copyright headers are missing and I think we should add a timeout to the workflow in case it gets stuck somewhere so that it does not run for forever. Once that is addressed you will get my approval.

The creation of the appwrite instance looks very good and will be very useful.

Regarding the integration test example, we might change that in a future sprint to cypress which is specialized for end-to-end-, unit- and integration-tests of websites. But that is just a comment. The example however is more then fine :D

.github/workflows/run_integration_tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: install and configure appwrite cli
+        working-directory: ./backend/setup_appwrite
+        run: |
+          ls
+          curl -sL https://appwrite.io/cli/install.sh | bash
+          ./init-cli.sh
+          #- name: install appwrite functions
+          #working-directory: ./backend/setup_appwrite
+          #run: |
+          #./create-set-wallet-function.sh
```
dominic-heil, line 33: Is there a reason why this is commeted out?

backend/setup_appwrite/init-cli.sh
```diff
@@ -0,0 +1,11 @@
...
+key=$(cat api-key.txt)
+project=$(cat project-id.txt)
+echo "$key"
+echo "$project"
+#appwrite client setEndpoint --endpoint="http://localhost/v1"
+#appwrite client setProject --project="$project"
+#appwrite client setKey --key="$key"
+#appwrite client setSelfSigned --value=true
+#appwrite client setLocale --locale="en-US"
+appwrite init --endpoint="http://localhost/v1" --project="$project" --key="$key" --locale="en-US"
```
steinkohl, line 11: Shouldn't be locale "de-DE" ? I think this also affects timestamps

steinkohl (approved):
Nice work in general.
But I would prefer passwords to be stored in githubs env varaibles, even if they are only intended for testing. If somebody misuses this script to deploy their production server, they will have a very fun time.

dominic-heil (approved):
LGTM

tuilagio:
can this PR be merged? I need some of these changes for my issue ...

PR #140, opened 2021-12-10
Implement the Revert of timesOutReservations

JaPilgrim:
-Drops will get value "Reservation-Timeout": seconds between drop & where reservation has to be bought
-on "drop" the timeout is set to be X seconds after blocktime
-externally function (to be called by front/backend) triggers check for TimedOut Reservations and reverts reservation to the state, as if they had never been "dropped"

-Without checking "Optimization" for only certain number of exectutions of functions ( couple of thousands ) wouldnt compile due to 24kb contract size limit.
-Doesnt Verify at least partly due to said reason.

### DIFF
--- a/blockchain/contracts/FlatNFTtheWorld.sol
+++ b/blockchain/contracts/FlatNFTtheWorld.sol
@@ -1397,7 +1397,7 @@ contract NFTtheWorld {
                 nftOwnerships[_dropHash][nftIndex].nftName,
                 nftOwnerships[_dropHash][nftIndex].nftSymbol
             );
-            uint256 nftToken = tokenContract.mintNFT(uri, msg.sender);
+           
             nftOwnerships[_dropHash][nftIndex].owner.transfer(
                 nftOwnerships[_dropHash][nftIndex].weiPrice
             );
--- a/blockchain/contracts/FlatNFTtheWorld.sol
+++ b/blockchain/contracts/FlatNFTtheWorld.sol
@@ -672,7 +672,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         _name = name_;
         _symbol = symbol_;
     }
-
+    
     /**
      * @dev See {IERC165-supportsInterface}.
      */
@@ -737,7 +737,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
     function symbol() public view virtual override returns (string memory) {
         return _symbol;
     }
-
+    
     /**
      * @dev See {IERC721Metadata-tokenURI}.
      */
@@ -1264,7 +1264,7 @@ contract NFTtheWorld {
 
     mapping(address => mapping(uint256 => string[]))
         private nftReservationInformationOfUsers;
-    mapping(address => uint256[]) private nftAssetsInformationOfUsers;
+    
 
     // Used to track which addresses have joined the drop
     mapping(uint256 => address[]) private joinedUsers;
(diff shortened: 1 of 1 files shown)


steinkohl:
I think that should be fine - I will test it tomorrow.
But haven't we decided to use a list + counter starting from 0 instead of the _drophash hashmap ? 
Syntactically the code should work in both cases, but the naming would be irritating

blockchain/contracts/FlatNFTtheWorld.sol
```diff
@@ -1262,7 +1264,7 @@ contract NFTtheWorld {
 
     mapping(address => mapping(uint256 => string[]))
         private nftReservationInformationOfUsers;
-    mapping(address => uint256[]) public nftAssetsInformationOfUsers;
+    mapping(address => uint256[]) private nftAssetsInformationOfUsers;
```
BerinikeTech, line 1267: If this is not needed anymore @steinkohl, then we could get rid of it. What do you think? We are getting this information directly from the wallet now, right?

blockchain/contracts/FlatNFTtheWorld.sol
```diff
@@ -1279,6 +1281,7 @@ contract NFTtheWorld {
         uint256 _dropTime,
         string[] memory _uris,
         uint256 _weiPrice,
+        uint256 _reservationTimeoutSeconds,
```
BerinikeTech, line 1284: When you change the functions, please also remember to update the documentation :)

blockchain/contracts/FlatNFTtheWorld.sol
```diff
@@ -1399,6 +1405,29 @@ contract NFTtheWorld {
...
+                nftOwnerships[_dropHash][i].reservedUntil =0;
+                reservedNFTsCount[_dropHash] -= nftReservations[nftOwnerships[_dropHash][i].reservedFor][_dropHash];
+                nftReservations[nftOwnerships[_dropHash][i].reservedFor][_dropHash] = 0;
+                nftOwnerships[_dropHash][i].reservedFor=nftOwnerships[_dropHash][i].owner;
+                availableNFTs[_dropHash].push(nftOwnerships[_dropHash][i].uri);
+                reservationsReverted++;
+            }
+
+        }
+        return reservationsReverted;
```
BerinikeTech, line 1428: I think if this is not needed for anything we should get rid of it

BerinikeTech (changes requested):

blockchain/contracts/FlatNFTtheWorld.sol
```diff
@@ -1399,6 +1405,29 @@ contract NFTtheWorld {
         }
     }
 
+
+    // to be called automatically from backend
+    // checks weither reservation has timed out & if so, if reservedFor != owner, meaning it wasnt bought
+    // if so, reinstate as if Drop was executed but NFT wasnt reserved 
+    function revertTimedoutReservations(uint256 _dropHash) public returns (uint256){
```
dominic-heil, line 1412: Maybe we should make this function executable only by admins so that the admins need to pay the fee and not some random person?

blockchain/contracts/FlatNFTtheWorld.sol
steinkohl: Yes, I overlooked that. Since we can fetch the owner directly from the NFTs, this is no longer necessary.

blockchain/contracts/FlatNFTtheWorld.sol
steinkohl: Yes your right, but I wouldn't do so.
One Senario: 
A company plans a drop in the future, but meanwhile goes bankrupt.
Then you are able to cheat and reserve many NTFs (with different wallets).
After the drop you only pay one requested batch, all other bought NFTs got lost / locked.
The few NFTs you've got would have gained rarity and value.

blockchain/contracts/FlatNFTtheWorld.sol
JaPilgrim: I think if we were to call this function, getting to know how many NFTs have been put back to distribute would be beneficial information. But isnt needed for now

blockchain/contracts/FlatNFTtheWorld.sol
```diff
@@ -1264,7 +1264,7 @@ contract NFTtheWorld {
 
     mapping(address => mapping(uint256 => string[]))
         private nftReservationInformationOfUsers;
-    mapping(address => uint256[]) private nftAssetsInformationOfUsers;
```
BerinikeTech: Please get rid of `nftAssetsInformationOfUsers[msg.sender].push(nftToken);` in `buyNFT` then as well :)

BerinikeTech (changes requested):

BerinikeTech (approved):
LGTM

PR #142, opened 2021-12-14
Closes #111 - Add landing page design

### LINKED ISSUE #111: Design countdown clocks and list of next NFT drops on the landing page
## User story
1. As a developer
2. I need an extended landing page design for a backwards running ticker/clock together with a list of public NFT-drops on the landing page
3. So that we can start implementing the design for André

## Acceptance criteria
* The current design is extended so that there is a vertical list of one or multiple horizontally displayed NFT drops.
* each simulatenously announced NFT drop has its digital clock which shows dd/hh/mm/ss time until release date.
* Each NFT Drop shows some short Drop information like name, creator, Shortened description.
* A list of NFT rarities contained in the NFT drop (with an example image for each rarity) can be displayed above or below a list of example NFTs.
* André had the optional idea of a background animation, see http://safemoon.net/
* Design should also be done for small screens.
* It may take inspiration from  websites as linked on this platform: https://newdrops.io/ .

## Definition of done
* Code is in the main branch

krisutofu:
Here is the design update for landing page. It includes additional interactivity.

Please see the comments in the Figma project which also explain how the design technically behaves:

https://www.figma.com/file/FRqpUl9lXzAZqG02BHk626/Home_page_base?node-id=0%3A1

I'm yet too much noob for how to use Figma properly so I used comments and no components. I was extremely slow or inefficiently working with it. Every change costs so much time, I practically worked 4 working days on this.

I don't know why the small screen version was exported so large:

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -1,0 +1,10 @@
+links for the Figma projects:
+
+* nft-the-world-landing2.fig
+
+  https://www.figma.com/file/FRqpUl9lXzAZqG02BHk626/Home_page_base?node-id=0%3A1
+
+* nft-the-world-mockups.fig
+
+  https://www.figma.com/file/8kdPSVa0kG9Bl5xVdLpydr/NFT-The-world!?node-id=19%3A81
+


krisutofu:
I also considered of all your last comments, @NickZecamp . Just the footer-banner, I thought, would be better to leave with the colour of the background because its position is fixed.

tuilagio:
looks good

tuilagio (approved):

krisutofu:
@NickZecamp I changed that loudspeaker to an eye. Is there some more things for this current design?

NickZecamp (approved):
Everything looks good to me 👌

Just discovered that you are using unicode characters in your icons. The problem is that they can be displayed in different ways and do not always follow the same style. What do you think if we replace them with images?

krisutofu:
> Everything looks good to me ok_hand
> 
> Just discovered that you are using unicode characters in your icons. The problem is that they can be displayed in different ways and do not always follow the same style. What do you think if we replace them with images?

Yes, that's okay. I didn't know about this problem. Could you help me with that? Maybe I should replace some of my cards and the input field with your field?

krisutofu:
> Everything looks good to me ok_hand
> 
> Just discovered that you are using unicode characters in your icons. The problem is that they can be displayed in different ways and do not always follow the same style. What do you think if we replace them with images?

I added a Figma commment there which explains to use images. It's too much for me now to change it, so I hope this sufffices.

PR #143, opened 2021-12-15
Follow up announcement

tuilagio:
This PR:
- [x] Fixed date. In the example collection python code, datetime resolution was "second". JS on the FE however expected "milisecond" resolution. Fixed by forcing second resolution on everywhere.
- [x]  Fix color, spacing.
- [x]  Change the `Edit` and `Delete` buttons to new style.
- [x]  Code overhaul.
- [x] Add fake images for announcements on front page. Main page will need another mockup and followup frontend+backend ticket.

<p align="center">

</p>

<p align="center">

</p>

Not good:
❌   Announcement sidebar on FAQ page is broken. Since there are lots of configurations in the FAQ code, I would rather let @krisutofu or @dominic-heil take a look. `frontend/src/pages/LandingPage.js` works just fine!
❌   The divider under announcement on sidebar is a bit longer than wanted. The positioning and containing is a mess.
❌   Announcement is somewhat responsive. Need more work though.
❌   `Readmore` is unsuccessful. There are solutions on the internet, however the effort that was put into integrating to our code base has already been too much without success due to some performance issue (not enough experiment with FE). @dominic-heil  feel free if you want to fix it.


krisutofu:
Nice that you also implemented some changes for my proposed design :-) . The links still could use some underlining and highlighting but maybe the underlining is not strictly needed.

The link texts should be PT Sans instead of Noto Sans font.

The images miss rounded edges as style property.

I'll check it out and see if something can be fixed.

Floinko (approved):

krisutofu (changes requested):
Changes are still needed. The images don't work correctly. Also my previously mentioned changes should be considered. I'll make some suggestions to fix the problems since I got it running now.

frontend/src/pages/AnnouncementPage.js
```diff
@@ -119,88 +130,163 @@ function AnnouncementEntry({
...
+							<div
+								style={{
+									backgroundImage: `url(${randomPhoto})`,
+									backgroundSize: "100%",
+									backgroundRepeat: "no-repeat",
+									backgroundPosition: "center",
+									width: "126px",
+									height: "100px",
+								}}
+							/>
```
krisutofu, line 161: I recommend writing this into a single line, and I would recommend outlining the style anyways for better maintainability but Dominic probably would not care here ;-) .

This is how the style should be:
```suggestion
							<div style={{ backgroundImage: `url(${randomPhoto})`, backgroundSize: "cover", backgroundRepeat: "no-repeat", backgroundPosition: "center", width: "126px", height: "100px", border-radius: "10px", }} />
```

frontend/src/pages/AnnouncementPage.js
```diff
@@ -119,88 +130,163 @@ function AnnouncementEntry({
...
+		fontFamily: "Noto Sans", fontSize: "11px", fontStyle: "normal",
+		fontWeight: "medium", opacity: 0.4
 	};
-	const contentStyle = { fontFamily: "Noto Sans", fontSize: "12px", fontStyle: "normal", fontWeight: "medium" };
+	const contentStyle = {
+		fontFamily: "Noto Sans", fontSize: "12px",
+		fontStyle: "normal", fontWeight: "medium",
+		color: "rgba(255, 255, 255, 0.81)"
+	};
+	const randomPhoto = Photos[Math.floor(Math.random() * Photos.length)];
```
krisutofu, line 144: Please remove the randomness here. When I change the screen size it will randomly change images which is not a good thing for a product presentation.

krisutofu: I did this:
```suggestion
	const announcePhoto = photos[parseInt(announcement.$id, 16) % photos.length];
```
I changed the names to `photos` and `announcePhoto` (via Renaming feature in VSCode).

krisutofu:
Is it okay, if I add some changes to this or a new branch?

frontend/src/pages/AnnouncementPage.js
```diff
@@ -119,88 +130,163 @@ function AnnouncementEntry({
...
+						</Link>
+					</Grid>
+					<Grid
+						item xs
+						sx={{
+							height: "112px", borderBottom: 1, pl: 1,
+							borderColor: "rgba(255, 255, 255, 0.1)", mb: 2
+						}}
+					>
+						<Typography style={titleStyle} sx={limitLines} variant="h5">
```
krisutofu, line 171: "Typography" is not right. It needs to change to HeaderTypography and ParagraphTypography respectively.

krisutofu (approved):

krisutofu:
I also removed the duplicate PDFs outside of the Deliverables folder.

frontend/src/pages/AnnouncementPage.js
```diff
@@ -10,21 +10,27 @@ import useChangeRoute from "../hooks/useChangeRoute";
...
+import examplePhoto4 from "../assets/img/announcementPhotoExamples/anouncement_4.png";
+import examplePhoto5 from "../assets/img/announcementPhotoExamples/anouncement_5.png";
+
+const photos = [examplePhoto1, examplePhoto2, examplePhoto3, examplePhoto4, examplePhoto5];
 
 function InputFields({ defaultTitle, defaultContent, titleComponenId, contentComponenId }) {
 	// TODO: formated text support
 	// ref: https://mui.com/components/text-fields/#integration-with-3rd-party-input-libraries
-	return <Grid sx={{ mt: 3, mb: 3 }}>
+	return <Grid sx={{ mt: 3, marginBottom: 3 }}>
```
krisutofu, line 33: I know, this change is actually unnecessary, but I forgot to undo it. I hope, this is okay.

krisutofu (approved):
If Dominic knows, why it needs 24 columns and doesn't work with 12, I'd be glad to know. I'm not taking more energy now to make this work.

tuilagio:
> Changes are still needed. The images don't work correctly. Also my previously mentioned small things should be considered. I'll try to help you to fix the problems since I got it running now.
> 
> The "Read more." only needs to point to the announcement page. What was the difficulty? Only Performance? Functionality and code quality matter over performance, particularly for an MVP.

The problem is doing it dynamically is not easy. Say you have a box 400x200 with known font and size. Now you can calculate to truncate a long text to fit into that box nicely, and add `readme` link to the end of the text. However, doing that for a responsive box would require more works and CSS and JS is not playing well for this purpose.

PR #144, opened 2021-12-15
134 setup cypress for frontend tests

dominic-heil:
This PR adds Cypress integration components tests that run independent of an AppWrite server (e.g. by using mocked data) as well as Cypress integration system tests. These tests are run on every push using a new and an edited workflow.

I tried to figure out code coverage for a few hours but did not get anything working yet. In theory it is possible but it is not trivial for me. I think we should create a follow up story for that if needed.


Floinko (approved):

steinkohl (commented):
@Floinko cloud you run it on your machine? I unfortunately can't at the moment. 
If it does work on your machine -> LGTM!
Nice code style, well documented and job/action logs are promising too!

frontend/cypress_system_tests.json
```diff
@@ -0,0 +1,13 @@
+{
+  "baseUrl": "http://localhost:3000",
+  "viewportWidth": 1920,
+  "viewportHeight": 1080,
+  "component" : {},
+  "integrationFolder": "cypress/system",
+  "pluginsFile": false,
+  "env": {
+    "APP_KEY": "f9c23feb7e5e70217a40516490d6ed3572c863f27716e183dbbc37d7d10dfb928434acea06a05637bfb7f294652336cb08789866e57ce41f81856bf15bd3c8f0dc646f32ab7fb04f6b437850941d0f9beefd3aa95097f06c6a5cd6e567594a1fc023d529b217ac33cb266124de2647a6bd0666acef34ca8eba4332e56f64a13a",
+    "APP_ENDPOINT": "http://localhost:81/v1",
```
tuilagio, line 10: is the port 81 the correct one?

dominic-heil: 81 was just a default value. As a usually have something else running on port 80, I need to use another port.

tuilagio (approved):

PR #146, opened 2021-12-15
NFT Drop page mockup added

NickZecamp:
Join drop
Join drop - Mobile
Created a mockup the Join drop page


krisutofu:
Hey, this is good! Your design is better than the one of mine.

Do you think, we could make a carousel or sidescroll component out of NFT drop list?

krisutofu:
It's also very fine to use the counter. I think, could be useful to have a date + a counter but this is up to your decision ;-) .

If you use the counter value, I rather would name it "Remaining Time" or "Time Left" instead of "Date" otherwise this is nice.

krisutofu (approved):

PR #147, opened 2021-12-16
96 create new drop

dominic-heil:
This PR (draft) implements a page where an admin can create a new drop. As I do not have admin access to any of our new contracts, I only tested it up until the point where I sent MetaMask the data which then displays that the transaction will fail (as it can pre-compute that I am not an admin). Once I am admin of a contract this will turn into a real PR. Other than that single component, everything was tested by me.

Remember that this story is not about perfect design but about the functionality of the page.

Some screenshots:
image
image

image
image
image


dominic-heil:
Just tested it properly with our most recent contract and me being an admin. Should be reviewable and mergable.

frontend/src/pages/CreateDropPage.js
```diff
@@ -0,0 +1,234 @@
...
+						label="Drop date & time"
+						value={newDropTimeDate}
+						onChange={(newValue) => {
+							setNewDropTimeDate(newValue);
+						}}
+					/>
+				</LocalizationProvider>
+
+				<ParagraphTypography style={{ paddingTop: "20px" }}>Enter the URIs of the new NFTs. Enter exactly one URI per line. The number of URIs determines the number of NFTs in the drop. </ParagraphTypography>
+				<TextField multiline rows={20} sx = {{ ...inputFieldStyle, paddingBottom: "20px"  }} margin="normal" required fullWidth id="nftUriList" label="NFT image URIs" name="nftUriList" autoFocus />
```
tuilagio, line 119: is this text field `NFT image URIs` required? I could create a drop although it was left empty. The Asterisk symbol is displayed in the UI

dominic-heil: Good catch. I added a condition in the formatting function to catch this.

tuilagio (approved):

BerinikeTech (approved):
LGTM

PR #153, opened 2021-12-16
Update Landing Page Design (2)

krisutofu:
2nd chance for you to review the PR. I did some changes.

The changes now include also the updated NFT Drop page design.

Updated figma project export:
https://www.figma.com/file/FRqpUl9lXzAZqG02BHk626/Home_page_base?node-id=0%3A1


dominic-heil:
In my opinion the NFT drop cards are way to stretched. The ratio of the cards is super weird, like 7:25 or so. Websites like openseas use something like 2:3. Maybe we want to make them like 1.5 times as wide? Then all the stuff in the cards also wouldn't look so crowded. @NickZecamp do you think the card ratio is really fitting?

NickZecamp:
> In my opinion the NFT drop cards are way to stretched. The ratio of the cards is super weird, like 7:25 or so. Websites like openseas use something like 2:3. Maybe we want to make them like 1.5 times as wide? Then all the stuff in the cards also wouldn't look so crowded. @NickZecamp do you think the card ratio is really fitting?

Yes, let's make them wider. 

We need to agree on the width of the viewport. Opensea has no max-width. We can do it on full screen too, or at least make it wider than now.

krisutofu:
> In my opinion the NFT drop cards are way to stretched. The ratio of the cards is super weird, like 7:25 or so. Websites like openseas use something like 2:3. Maybe we want to make them like 1.5 times as wide? Then all the stuff in the cards also wouldn't look so crowded. @NickZecamp do you think the card ratio is really fitting?

Sounds reasonable. You could just do like you supposed and it would be okay. I think deviating from the design ratios are not dramatic. I only wanted to fit all three cards into one screen.

dominic-heil:
Also, I don't know how much sense it makes to display the individual NFT Cards/images on the landing page already. The user would need to scroll up and down multiple times as the content does usually not fit one one page which does not feel too natural for a landing page in my opinion. I think it would be sufficient to just redirect to the individual drop page if the user clicks on a nft drop card. Otherwise the landing page might also feel too crowded. But that might just be my taste.

krisutofu:
> Also, I don't know how much sense it makes to display the individual NFT Cards/images on the landing page already. The user would need to scroll up and down multiple times as the content does usually not fit one one page which does not feel too natural for a landing page in my opinion. I think it would be sufficient to just redirect to the individual drop page if the user clicks on a nft drop card. Otherwise the landing page might also feel too crowded. But that might just be my taste.

I had a similar thought about the scrolling but I didn't know how to change it. I would keep the listing of NFT-Drops (due to easier overview) but put away the NFT-cards that are shown under the drop list, when a NFT-drop entry is selected. I actually would like it, if I could see 3 example NFT images for each NFT-Drop entry so that I can more easily see whether I want to know more about this NFT-drop. These 3 images don't need to be cards, just simple little images shown with each NFT-Drop entry which can be enlarged on screen when clicked. I could try to adapt the Figma design.

As for the NFT-card container below the NFT-drop list: you can integrate the missing parts into the NFT drop page as you suggest.

The lost feature is the ability to quickly switch the previewed drop content when the banner's side arrow at the left and right is clicked. But it's probably sufficient, to display the banner on the NFT-drop page and the side arrow of the banner would just refer to the next drop page when clicked. It would mean bigger load times but nevermind.

dominic-heil:
> I had a similar thought about the scrolling but I didn't know how to change it. I would keep the listing of NFT-Drops (due to easier overview) but put away the NFT-cards that are shown under the drop list, when a NFT-drop entry is selected. I actually would like it, if I could see 3 example NFT images for each NFT-Drop entry so that I can more easily see whether I want to know more about this NFT-drop. These 3 images don't need to be cards, just simple little images shown with each NFT-Drop entry which can be enlarged on screen when clicked. I could try to adapt the Figma design.

How about a tooltip with some sample images? E.g. you hower over a drop card and then above (or below) a tooltip is displayed in which an atomatic carousel switches through some sample images. Just an idea and everything would be in view at the same time.

krisutofu:
> How about a tooltip with some sample images? E.g. you hower over a drop card and then above (or below) a tooltip is displayed in which an atomatic carousel switches through some sample images. Just an idea and everything would be in view at the same time.

Sounds great! I like tooltips a lot and it's about what I would imagine.

I changed the design now.

The small screen design still mixes the landing page design with the drop page design. You can ignore (leave out) the side-stepping NFT-Drop-selection component there.

For phones on the landing page, it might be more useful to have smaller cards with less information for each NFT-drop, maybe just the backwards clock text. And more information can be displayed when it is clicked on.

dominic-heil (approved):
LGTM now!

krisutofu:
Would we like to merge this one? Or not?
@NickZecamp 
Do you agree to the current figma project?

The PR also removes my own Figma project file in order to remove the effort of maintaining an extra file in the Repo. All Figma project links are in the readme file.

NickZecamp (approved):
Yes, everything looks good!

Maybe my eye is catching on to the arrows buttons that scroll through the list of items. I usually prefer not to have more than one scrollbar on a site or carousels, arrow buttons. Unless it's a product pics or a banner. I would make the entire list of elements with dynamic loading of content when scrolling. But this is not always suitable for every project, so I do not have a strong opinion on this.

krisutofu:
> Yes, everything looks good!
> 
> Maybe my eye is catching on to the arrows buttons that scroll through the list of items. I usually prefer not to have more than one scrollbar on a site or carousels, arrow buttons. Unless it's a product pics or a banner. I would make the entire list of elements with dynamic loading of content when scrolling. But this is not always suitable for every project, so I do not have a strong opinion on this.

Thanks.

Okay. This would work well the horizontal "card"s (the vertical list). Dynamic loading could be good for usability with a lot of NFT Drops available but has no priority in my point of view. I don't know effort for doing that in a good way. If the arrow buttons should be very bad, we just could disable them in the code but for now, Dominic implemented them nicely.

PR #155, opened 2021-12-20
Close #132 - Add React component for restricted access

### LINKED ISSUE #132: Add "Partners" team to frontend
## User story
1. As a partner of the system 
2. I want to have the role of a partner on the platform
3. So that I gain special permissions

## Acceptance criteria
* There is an frontend component which only allows "Partners" and / or Admins to view its content.
* "Partners" role is shown on the profile page.

## Definition of done
* Code is in the main branch
* (code can be tested manually or automatically?)

krisutofu:
Replaced the area folder with a generic RestrictedAccess component which also permits partner access.

This should adapt the frontend to the new partner group.

A tiny test is added to the App.js page.


frontend/src/components/RestrictedArea.js
```diff
@@ -0,0 +1,87 @@
...
+
+const loadingMessage = <ErrorMessage>
+	Loading...
+</ErrorMessage>;
+
+const noMemberAccessMessage = <ErrorMessage>
+	You are trying to access a restricted area. If you believe you should have access to this area, please contact an admin directly.
+</ErrorMessage>;
+
+const adminTeamName = "Admins";
```
tuilagio, line 30: can we put these constants into a `config` file to reuse later?

krisutofu: Good idea, I put it into `utils/config.js`.

frontend/src/components/RestrictedArea.js
```diff
@@ -0,0 +1,85 @@
...
+	const updateMemberState = async () => {
+		if (!user || !Array.isArray(teams)) return;
+	
+		const teamsObject = await appwriteApi.listTeams();
+		const isTeamMember = arrayIntersection(teamsObject.teams.map(team => team.name), teams);
+		setUserIsTeamMember(isTeamMember.length != 0);
+		setIsLoaded(true);
+	};
+
+	useEffect(updateMemberState, [user, teams]);
```
dominic-heil, line 66: I would just put the function in the useEffect immediately.

frontend/src/components/RestrictedArea.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Displays children only if the user is member of a specific team,
+ * otherwise wrapped components will not be shown.
+ * @param children wrapped components that should only be shown to admins
+ * @param user user object as maintained by the main application component.
+ * @param teams list of team names whose members can access `children`. Any logged in user if no array.
+ * @param elseIgnore flag indicating whether error messages are suppressed.
+ * @returns {JSX.Element}
+ */
+export default function RestrictedArea({ user, teams, children, elseIgnore }) {
```
dominic-heil, line 52: I dont think `elseIgnore` is an intuitive name. Took me quite some time to understand the meaning of the variable. Maybe change it to something like `suppresErrorMessages` and add a small example in the JSDoc to understand what to use the variable for?

dominic-heil (changes requested):
One small change request, rest LGTM.

frontend/src/components/RestrictedArea.js
krisutofu: I know, that you guys like to do this but I think, nesting anonymous lambdas is worse for readability. It's okay for Python lambdas, they are one-liners. I think, inlining a lot of code is rather messy.
This way, people know, that the member state is updated, when the `user` and `teams` variables change.

Would it be okay for you to leave it this way?

tuilagio:
can we already merge this one? looks ok to me

dominic-heil (approved):

PR #157, opened 2022-01-05
135 test action annouce collection

tuilagio:
- add integration test for announcement and wallets collection.


dominic-heil:
Is there any chance to write the created collection IDs, especially for the announcements, in a .txt file. I think when we create the collection, the returned document already contains the ID which would only require maybe three more lines to write it to a file. Then we can test stuff for the collection in our cypress tests as well =) Other than that, LGTM!

tuilagio:
> Is there any chance to write the created collection IDs, especially for the announcements, in a .txt file. I think when we create the collection, the returned document already contains the ID which would only require maybe three more lines to write it to a file. Then we can test stuff for the collection in our cypress tests as well =) Other than that, LGTM!

Sure thing, I can add that later today.

dominic-heil (approved):
LGTM! Well done

backend/database-collection-schemas/createAnnouncementCollection.py
```diff
@@ -26,44 +60,49 @@
...
+    )
+    print(createCollectionResult)
+    print(createCollectionResult["$id"])
+    with open("announcementCollectionID.txt", "w") as f:
+        print(createCollectionResult["$id"], file=f)
+except Exception as e:
+    print(e)
+    sys.exit(1)
 # Create some fake data
 data = [
```
krisutofu, line 107: better would have been
"announcementEntries"
or
"announcementTuples"

backend/database-collection-schemas/createWalletsCollection.py
```diff
@@ -16,26 +48,35 @@
...
+        ],
+    )
+
+    print(createCollectionResult)
+    print(createCollectionResult["$id"])
+    with open("walletID.txt", "w") as f:
+        print(createCollectionResult["$id"], file=f)
+except Exception as e:
+    print(e)
+    sys.exit(1)
```
krisutofu, line 80: This code duplication is just a little bit sad.

backend/python_init_script/requirements.txt
```diff
@@ -1,4 +1,4 @@
-appwrite
+appwrite==0.5.1
```
krisutofu, line 1: What was your reason to require a specific version of Appwrite? Security purposes?

krisutofu (approved):
Found no functional problems.

backend/python_init_script/requirements.txt
tuilagio: no, compatibility. `create_collection` i.e, requires extra `uid` in input parameters. I didn't want to deal with this, so lock the version in this script would be best fit.

backend/database-collection-schemas/createAnnouncementCollection.py
tuilagio: There could be failure message returned. In success case, this will return:
```json
{
    "$id": "5e5ea5c16897e",
    "$collection": "5e5ea5c15117e",
    "$read": [
        "role:all"
    ],
    "$write": [
        "user:608f9da25e7e1"
    ]
}
```
So I don't think there should be need for changing the var name here.

PR #159, opened 2022-01-11
Close #154 Add NFT Studio sketch

### LINKED ISSUE #154: Design a sketch for NFT creation (primitive NFT Studio)
## User story
1. As a frontend developer
2. I need to have an idea of what the NFT creation / NFT studio is about
3. So that I can start implementing the required NFT creation feature.

## Acceptance criteria
* The sketch is created in Figma online. If a new project is created, an online link is added to `assets/README.md`.
* It contains the ability to set/upload the custom contents of the NFT. The NFT's content is specified on that page. Yet, we only will support images as resource.
* The uploaded resources can be previewed individually.
* There is a UI element to preview the created NFT.
* All "NFT templates" that were created by the current user are listed somewhere and can be opened or deleted. Deletion is only possible when the associated NFT drop is not online, i.e. is not published as example NFT of an NFT drop.
* NFT drop ID is not required information. [Note: created NFTs will be assigned to drops and not the other way around]
* Every NFT template has a name. This is not an official name for the NFT.
* There is a way to access the NFT studio as an admin (or verified partner).
*
* _may_ but not must:
* It _may_ allow the resource view to have additional hierarchical structure via folders.
* It _may_ allow uploading zip-directories (or tar.gz whatever) which adds multiple files (with folders if applicable).
* An NFT drop ID and the NFT variety _may_ be specified in the page when a randomization feature is in the design and is enabled. All randomized NFT instances then would go into that NFT drop.
* 
* It _may_ allow for composing NFTs from multiple components. Each component has a way to be added, moved, resized or removed from the NFT preview image. In that case, there is a list of every component in the template. The currently highlighted list entry corresponds to the currently selected component.
* It _may_ allow for randomizing a component from multiple sources.
* A randomized component is a box instead of a specific image which uses a list of file name formats with wildcards instead of a single specific file name.
* If randomization is enabled, an NFT template also has a set of logic constraints that the randomization must satisfy. A logic constraint is a logic clause that uses names of randomized boxes and file name formats, relates them via non-/assignment. The set of all constraints are ANDed.
* The NFT template _may_ have a set of rarities which define for each image a probability to appear. The sum of probabilities don't need to be 100% but not surpass it. [This allows the probability that a randomized box remains empty during random generation.]
* When randomized boxes are used, an example NFT can be rendered with a render button. The example is either a valid random NFT or a specific one that is composed from highlighted resources in the resources panel.

## Definition of done
* Code is in the main branch
* (code can be tested manually or automatically?)

krisutofu:
I added a Figma project which sketches the NFT Studio that I envisioned.

NOTE: most of this design is optional to implement, more like an idea how it could be developed after our project.

I found it useful to include an updated design for the NFT Drop creation page since I scrapped the Drop Workshop idea due to time. With a minimal approach, NFT creation equates to NFT Drop creation.

It is used to define NFT templates from a set of random resource images.
I also scrapped the NFT Drop Workshop idea and tried to integrate the minimum necessary functionality into the studio page which includes some NFT Drop information of the NFT template.

The sketch needs to be finalized by replacing some text with input or text UI elements but the concept is finished.
I tried to satisfy all optional user story criteria. But a lot of it is likely out of scope for our project due to time.

It is here:

https://www.figma.com/file/EJfBJHqdGNRExHcvAL2KTV/NFT-Studio?node-id=0%3A1


dominic-heil (approved):
Looks good as a concpet for the future. However, as we already discussed, I think it is unrealisic to implement this in a timely manner now.

NickZecamp (approved):
Everything looks nice to me!

krisutofu:
> Looks good as a concpet for the future. However, as we already discussed, I think it is unrealisic to implement this in a timely manner now.

Okay, let's see, if there are capabilities at the end of the project. The current NFT Drop creation page doesn't look too bad, honestly. The usability isn't so bad either. Rather minimum information is only required to create the NFT Drop.

PR #195, opened 2022-01-20
Add new profile design implementation

krisutofu:
I implemented the design for our user profile.

The small screen version is also included.

All of the older functionality is taken over to this new profile.

The functionality is not complete and I don't know, if everything is working like before. Would be very useful here to have regression tests. The Change-Password page is now integrated into the profile.
invalid_credentials
unverified_profile
verified_email


frontend/src/pages/Profile.js
```diff
@@ -28,60 +111,245 @@ import { textColor } from "../assets/jss/colorPalette";
...
-					})}
-				>Logout</Button>
-			</Grid>
-		</Grid>
-
-	</CenterFlexBoxMedium>;
-}
\ No newline at end of file
+	const render = () => (<Box component="form" onSubmit={checkAndSaveProfile} >
+		<EmailStatusBanner isVerified={isEmailVerified}/>
```
dominic-heil, line 119: For perfection, some margin is missing in the mobile version 
image

frontend/src/pages/Profile.js
```diff
@@ -28,60 +111,245 @@ import { textColor } from "../assets/jss/colorPalette";
...
+
+		return (isVerified)?
+			(<div style={{ background: "#D7FFDD", ...emailStatusBannerStyle, display: "flex", alignItems: "center", }}>
+				<GreenCheck/>
+
+				<Margin width="13px"/>
+
+				<ParagraphTypography style={{ display: "inline", fontWeight: "400", fontSize: "15px", color: "#456548", }}>
+					Your email was successfully confirmed
+				</ParagraphTypography>
```
dominic-heil, line 231: Shouldn't this be shown only once and not every time?

dominic-heil (approved):
LGTM. Only thing I maybe would additionally change would be the vertical centering of texts on the very right. Well done

dominic-heil:
Maybe we should also add a seperat button for changing the password if you want to change the user name but not the password instead of having a button that changes all?

frontend/src/pages/Profile.js
krisutofu: > Maybe we should also add a seperat button for changing the password if you want to change the user name but not the password instead of having a button that changes all?

Very good idea, or maybe a "reset" button to reset a profile setting to it's last saved value. I should do more work here.

krisutofu: Yes, this is the missing margin that I gave up to get some sleep ;-) .
The reordering logic got more complicated than it should be. I'll try to improve it.

krisutofu:
Could someone verify whether the Wallet works? Changing Password works for me.

When I connect the wallet, I get an AppwriteException that I cannot execute a task for role "" because my role is "member". Seems like it expects wrong permissions on my side.

dominic-heil:
The wallet thingy worked for me earlier.

frontend/src/pages/Profile.js
dominic-heil: I think we do not need to display the information at all if the mail is verified as this info is also displayed on the profile page anyways.

Floinko (changes requested):
I would at least leave out the profile picture. Currently no such feature is planned and i think having functionality in here that is not implemented would be a step back. Also i would only have the button for changing mail / username if that really works. I think in the mockups, those text fields were meant to be disabled inputs (for optic reasons) and not really editable fields.

krisutofu:
> I would at least leave out the profile picture. Currently no such feature is planned and i think having functionality in here that is not implemented would be a step back. Also i would only have the button for changing mail / username if that really works. I think in the mockups, those text fields were meant to be disabled inputs (for optic reasons) and not really editable fields.

However, it would not be a problem to just disable the textfields. Very small effort, wait a moment.

The password and wallet thing works like before. And for the profile picture, it belongs to the design, so I think, it wouldn't be fitting to cut it out (even though it would be very simple to do so.

krisutofu:
> I agree that it matches the design. However in the meeting last week i brought up my concerns about the design itself. I had to leave on time this week so i don't know what you discussed about this ticket, but i still think that we should not merge parts of features that we are not planning to finish anyways.

It took me quite a while to figure out how to change the font color of a disabled TextField. There was no solution online so I had to find it out myself. Terrible.

Would it be okay to just hide it in the last sprint, if not implemented? The profile picture actually was not the easiest to get right. If someone wants to recreate the design later, this will take just another effort.

krisutofu:
Let's merge fast so that Que can continue working with the NFT info page which I refactored here a bit.

tuilagio:
Overall, looks good. Some points:

- The wallet area needs more work.
image
- I tried change username. Didn't work for me.

However, these can be fixed later. No time for perfection.

tuilagio (approved):

PR #196, opened 2022-01-20
167 integrate real data from database in frontend

dominic-heil:
This PR replaces the mocked data on the landing page and the drop page. The content off the users collections is not yet replaced as there are some contract funtionalities missing for that atm.

image
image


frontend/src/components/NftDropCardStructuredList.js
```diff
@@ -49,9 +49,8 @@ export default function NftDropCardStructuredList({ nftDataArray }) {
 			orderType = "DESC";
 			break;
 		case "hot":
-			// TODO: order by left NFTs
-			orderField = "drop_id";
-			orderType = "ASC";
+			orderField = "drop_reserved";
```
steinkohl, line 52: Well, maybe the "hot" ordering should be based on a normalized value, otherwise a drop with 3/10000 would be hotter then a limited drop with 2/5 NFTs already reserved

steinkohl (approved):
LGTM

frontend/src/components/NftDropCardStructuredList.js
krisutofu: My original idea of "hot" was the current "speed" of sold NFTs. It's difficult to compute a value without having the time therefore I proposed a "hot" similar to eBays "only X left", but I admit, the idea is not perfect :-/ .

steinkohl: nothing against this idea, the "problem" here is only that isn't ordered by percent but by absolute counts

tuilagio (approved):

PR #198, opened 2022-01-23
184 reserved nfts logic

BerinikeTech:


blockchain/contracts/MainContract.sol
```diff
@@ -121,13 +123,13 @@ contract NFTtheWorld {
...
-        reservedNFTsCount[dropHash] = 0;
     }
 
     // This function lets a user join a drop by specifying the number of NFTs she would like to reserve.
     function joinDrop(uint256 _numberOfNFTs, uint256 _dropHash) public {
         require(
-            reservedNFTsCount[_dropHash] != availableNFTs[_dropHash].length,
+            dropData[_dropHash].reservedCount !=
+                availableNFTs[_dropHash].length,
             "Cannot join the drop anymore."
```
JaPilgrim, line 133: As now "availableNFTs" gets diminished (call to remove() ) within the function it doesnt represent the number of total NFTs available anymore, but the number of currently available NFTs. This throws error if at excactly 50% reserved NFTs of drop. Either:
- work with numberofURIs
- dont remove from availableNFTs at all

JaPilgrim (changes requested):
Requested Changes, Rest looks good & is working

dominic-heil (approved):
LGTM

steinkohl (approved):
Tested in remix - LGTM

JaPilgrim (changes requested):
While Testing and executing "Buy" I get just an error message "Fail" 
image
Does it work for you all?

dominic-heil (changes requested):
Re-tested everything on the kovan network. Worked fine for me. However, the `mintedNFTContracts` call is very unpractical as you need to enter and index manually. It would be a lot better if the call could return a whole array of contracts.

JaPilgrim:
> > While Testing and executing "Buy" I get just an error message "Fail" image Does it work for you all?
> 
> @JaPilgrim Worked for me: https://kovan.etherscan.io/address/0x69779179BfF1429D89874882953B127e2e4C8eB2. Have you set a proper drop time and a large enough reservation timeout?

Yes I did set fitting droptime & timeout (also there was no error message, indicating the error doesnt result from not-reached time). 

However, when using the Deployed Version directly in Etherscan and not in Remix, "Buy" is working

JaPilgrim (approved):

dominic-heil (approved):
LGTM

PR #213, opened 2022-01-31
Real FAQ added

NickZecamp:
The following real questions and answers to the FAQ page have been added:

**What is an NFT?**
NFT or non-fungible token is a type of crypto asset that represents digital works of art and other collectibles into one-of-a-kind. Each NFT is unique and exists in a single copy, it cannot be divided, and all information about its author, buyer, and all transactions with it is securely stored in the blockchain. In other words, an NFT is a digital certificate attached to a unique object.

**Which blockchains does NFT world! support?**
NFT World! currently supports Ethereum blockchain.

**How to connect wallet?**
If you have not yet connected your wallet, you will be prompted to connect to the wallet when you try to join the drop or purchase an NFT token. Follow your wallet's instructions to complete the connection.

Also, you can do it in your Profile by clicking the Connect MetaMask Wallet button.

**Is connecting my wallet to NFT world! secure?**
When you connect your wallet, you give us the right to view your wallet addresses, but we cannot take any money from it.

When making any transaction, you will receive a notification in the pop-up window of your wallet, and you will need to confirm the transaction manually.

Thus, the connection to the wallet is secure, but you should still follow the general security rules. Such rules are, for example, to avoid using malicious sites and not to share your seed phrase with anyone.

**How can I join a drop?**
You can view and choose the drops you are interested in on the main page of the NFT World! After being redirected to the drop page, you will be able to view the NFT belonging to that drop. If you want to join the drop, then select the amount of NFT you would like to receive and click the Join Drop button.

You will be redirected to your wallet. Follow the instructions in your wallet to confirm the transaction.

**What are gas fees, and how much do I need to pay?**
The gas fees can be customized by a user with MetaMask. There you can set what you'd like to pay for the transaction, and lower gas fees mean slower transactions. MetaMask also says that transactions may not succeed if the fee is chosen too small.

**What is the NFT Collection, and where can I find it?**
The Collection shows the NFTs that you own. You can see your collection by clicking the NFT collection button on the Profile page.

**What is the NFT World?**
The NFT World is an open-source marketplace platform that can be used to sell and buy NFTs.
This solution allows small, medium-sized companies and enthusiasts to easily deploy their own NFT platform for non- or commercial purposes.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -3608,6 +3608,7 @@
         "jest-resolve": "^26.6.2",
         "jest-util": "^26.6.2",
         "jest-worker": "^26.6.2",
+        "node-notifier": "^8.0.0",
         "slash": "^3.0.0",
         "source-map": "^0.6.0",
         "string-length": "^4.0.1",
@@ -7808,6 +7809,7 @@
       "dependencies": {
         "anymatch": "~3.1.2",
         "braces": "~3.0.2",
+        "fsevents": "~2.3.2",
         "glob-parent": "~5.1.2",
         "is-binary-path": "~2.1.0",
         "is-glob": "~4.0.1",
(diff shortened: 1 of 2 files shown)


steinkohl (approved):
LGTM

krisutofu (changes requested):
Well done, only a few content related things that need changes.

I hope it's okay for you when I added these changes to the branch and some usability feature of being able to switch between single and multiple question mode.

PR #219, opened 2022-02-02
176 announcement images

krisutofu:
Adds an Editable Image component as generalized image component. It can load and save images from/to database (the editability is a parameter).

Screenshot from 2022-02-02 06-47-48
Screenshot from 2022-02-02 06-48-15

Storage API convenience functions are added.

It is integrated into the announcements and also could be easily integrated into the profile, working out of the box.

There are 3 major problems left:
- using the appwrite Web SDK createFile returns a "bad request". Opening files seems to work.
- Somehow, it tries to load images with ID 0 although it is undefined for all announcements. (happens after failed image upload, problem with EditableImage??)
- Image IDs are generated twice in pairs of two consecutive announcements. This happens because the state update – which sets the next imageID (if not available) – interferes with loading the next announcement, so that every odd-indexed announcement doesn't get the updated imageID number.

I tried to fix all the warnings in other unrelated components. Tiny amount of refactoring included.

Backend: "creator" and "imageID" are added as properties to the announcement database collection.

I didn't find to have broken existing announcement functionality.

*There are merge problems. Unrelated file additions need to be removed.*


frontend/src/api/appwriteApi.js
```diff
@@ -201,6 +195,45 @@ let api = {
...
+			console.error(`Cannot save image with ID ${fileID}, likely no write permissions.`);
+			return false;
+		}
+	},
+
+	removeImageFromDatabase(imageID) {
+		const fileID = this.imageToFileID(imageID);
+		return this.sdk.storage.deleteFile(fileID)
+			.catch(() => console.error(`Couldn't delete image with ID ${fileID}. Either doesn't exist or no proper permissions.`));
+  },
```
krisutofu, line 236:
```suggestion
	},
```

frontend/src/api/appwriteApi.js
```diff
@@ -201,6 +195,45 @@ let api = {
...
+			return true;
+		} catch(e) {
+			console.error(`Cannot save image with ID ${fileID}, likely no write permissions.`);
+			return false;
+		}
+	},
+
+	removeImageFromDatabase(imageID) {
+		const fileID = this.imageToFileID(imageID);
+		return this.sdk.storage.deleteFile(fileID)
```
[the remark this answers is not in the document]
krisutofu: Good remark!
I indeed removed the writePermissions from the file creation call when I made the PR.

Announcements can only be created/edited by admins at the moment (using `AdminArea` component) so that deleting announcements (or updating images) will also delete the old picture. I had "bad request" problems that's why I removed the writePermissions on file creation. I used the same way like Que to set the write Permissions. Maybe, it wasn't the problem.

frontend/src/api/appwriteApi.js
```diff
@@ -9,16 +9,10 @@ import { AppwriteServer, domainName, adminTeamName } from "../utils/config";
...
-	provider: () => {
-		if (api.sdk) {
-			return api.sdk;
-		}
-		let appwrite = new Appwrite();
-		appwrite.setEndpoint(AppwriteServer.endpoint).setProject(AppwriteServer.project);
-		api.sdk = appwrite;
-		return appwrite;
+	provider() {
+		return this.sdk;
```
dominic-heil, line 15: Why did you do this? The previous code was based on the official React example of AppWrite: https://github.com/appwrite/demo-todo-with-react/blob/main/src/api/api.js. By default, sdk stays null and no computation happens at all until provider() was called. The way you changed it sdk is always initialized, even if not needed at all. I think we should keep it the old way.

frontend/src/api/appwriteApi.js
```diff
@@ -201,6 +195,45 @@ let api = {
...
 
+	/**
+	 * Storage API
+	 **/
+	imageToFileID(imageID) {
+		return imageID && `img${imageID}`;
+	},
+
+	async loadImageFromDatabase(imageID) {
+		const fileID = this.imageToFileID(imageID);
```
dominic-heil, line 206: For consistency, we should use the same way of declaring functions throughout the api file imo.

frontend/src/api/appwriteApi.js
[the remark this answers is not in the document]
dominic-heil: The write permission should be set to `["team:{adminTeamID}"]` for sure.

dominic-heil (changes requested):
Haven't run in locally yet. Will do so later

frontend/src/api/appwriteApi.js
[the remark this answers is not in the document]
dominic-heil: Also, just because an area is protected by the `AdminArea` component in the frontend, you cannot expect that the user finds a way to access that area anyways (e.g. if someone knows how to read through our code in the browser which isn't too difficult). That is why any action in the admin area always also is protected by admin permissions through the backend.

frontend/src/api/appwriteApi.js
krisutofu: Is there a particular reason for not being initialized at the start?
It makes code more difficult to understand and the if-block is less efficient in the long-term.

Functionally, I didn't notice a difference. I found it more reasonable to directly initialize it at the start.

Well yes, the official Appwrite Code that I looked at doesn't look like good example code in general.

frontend/src/api/appwriteApi.js
krisutofu: Do you mean using `this` and not using the lambda expression syntax?

frontend/src/api/appwriteApi.js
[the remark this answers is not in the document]
krisutofu: Good point. I didn't exactly know, where which code is executed. Is it right that the functions are executed on server and the JSX is processed by the client? Or is everything run on client side?

In the most recent version in my other PR, I changed some things back due to the API overhead created by the AdminArea component. It rather makes sense for less-frequently updated content.

And if an attacker could potentially remove the AdminArea on client side, would these API calls have effect when used? Or rather not, because of lacking Admin rights?

frontend/src/App.js
```diff
@@ -43,15 +43,15 @@ function App() {
...
-	useEffect(() => {
-		appwriteApi.getAccount()
-			.then((r) => {
-				setIsLoaded(true);
-				setUser(r);
-			})
-			.catch(() => {
-				setIsLoaded(true);
-			});
+	useEffect(async () => {
```
krisutofu, line 46: Maybe passing an async lambda is not such a good idea. In the RestrictedArea Component, doing that gave a React state update warning.

frontend/src/api/appwriteApi.js
```diff
@@ -201,6 +195,45 @@ let api = {
...
+	async saveImageToDatabase(image, imageID) {
+		const fileID = this.imageToFileID(imageID);
+		
+		this.sdk.storage.deleteFile(fileID)
+			.catch(() => console.log(`No image to update. Create new image ${fileID}!`));
+
+		try {
+			const readPermissions = ["*"];
+			console.debug("save file with ID", fileID);
+			await this.sdk.storage.createFile(fileID, image, readPermissions);
```
krisutofu, line 224: Now I get, why it didn't work. They changed the interface between v0.11 and 0.12 and I used the newest one.

krisutofu: Does anyone know how to get the File ID for v0.11?

krisutofu:
Yesterday afternoon, it came to my mind how it should work. I throw away the stupid idea with progressively loading announcements, the image ID state updates and so on. I used the newest storage API which we don't use in our project. I changed it to v0.11. My self-generated image ID idea doesn't fit to that API anyways.

I didn't test the functionality of the EditableImage container.

PR #223, opened 2022-02-02
Update README.md

BerinikeTech:


blockchain/README.md
```diff
@@ -8,15 +8,15 @@
...
+* Feel free to test!
 
 
 ## Verifying & Publishing 
 * 1. After deploying contract(s) go to https://kovan.etherscan.io/ and look for contract address
 * 2. Hit tab "Contract" mid-page and just below click "Verify & Publish"
-* 3. Select: Solidity (Single file), Compiler Version v0.8.0 & respective license type & continue
-* 4. In next step, copy & paste contract & source code, and eventual Constructor Arguments & other setting (not needed atm)
-* 5. Hopefully find successfully generated ByteCode on next page
+* 3. Select: Solidity (Single file), Compiler Version v0.8.10 & respective license type & continue
```
JaPilgrim, line 17: But we are using 8.0 aren't we?
As it say: pragma solidity ^0.8.0; and also for verification I've always used 8.0.

blockchain/README.md
```diff
@@ -28,17 +28,28 @@
 
 ## Files 
 
-* truffle-config.js: Truffle configuration file for setting network information and other project-related settings.
+* truffle-config.js: Truffle configuration file for setting network information and other project-related settings like compiler type (we are using version 0.8.10 atm).
```
JaPilgrim, line 31: Same as above

blockchain/README.md
```diff
@@ -28,17 +28,28 @@
...
 
-* migrations/2_hello_world.js: Migration script for the HelloWorld contract. (Migration scripts are run in order, so the file beginning with 2 will be run after the file beginning with 1.)
+* migrations/2_deploy_token_factory.js: Migration script for the TokenFactory contract. (Migration scripts are run in order, so the file beginning with 2 will be run after the file beginning with 1.)
+
+* migrations/3_deploy_main.js: Migration script for the main contract.
+
+
+## Technical documentation
+
+Both contracts are currently deployed on the kovan ethereum testnet. We decided to use kovan since it is not as vulnerable to spam attacks as other alternatives and has been online long enough for having a rather large community and therefore a lot of support.
```
JaPilgrim, line 48: Maybe also add something like "can easily be deployed to ethereum mainnet or other testnets."

JaPilgrim (changes requested):
Suggestions like above, besidse it looks good!

blockchain/README.md
BerinikeTech: ^0.8.0 means >=0.8.0 and I always use 0.8.10 when verifying :D But can also change if you'd like me to :)

blockchain/README.md
BerinikeTech: done :)

blockchain/README.md
steinkohl: I would leave it like that, since minor changes are mostly security fixes and won't break anything

steinkohl (approved):
LGTM

blockchain/README.md
JaPilgrim: Ahh okay just thought that during verification it would still lead to an error if the minimal-versions were not matching, but if  thats not the case this is fine aswell

JaPilgrim (approved):
lgtm !

PR #234, opened 2022-02-02
Fix announcement issues

krisutofu:
There were multiple display bugs.

On changes or updates, it mainly wouldn't reload correctly while the action took place under the hood.

Did also minor refactorings here. The buggy RestrictedArea is updated as well (like in my other PR).

Unfortunately, Appwrite seems to be not working correctly. The announcements initially added via Python cannot be found by using the JavaScript API, even when loading those documents explicitly via their document ID.

One could also use Appwrite's API for listing database documents in a sorted way but the API seems to be unstable. It just was changed between v0.11 and v0.12 so a custom solution remains better in my opinion.

### DIFF
--- a/backend/database-collection-schemas/createAnnouncementCollection.py
+++ b/backend/database-collection-schemas/createAnnouncementCollection.py
@@ -94,6 +94,20 @@ try:
                 "required": True,
                 "array": False,
             },
+            {
+                "label": "creator",
+                "key": "creator",
+                "type": "text",
+                "required": True,
+                "array": False,
+            },
+            {
+                "label": "imageID",
+                "key": "imageID",
+                "type": "numeric",
+                "required": False,
+                "array": False,
+            },
         ],
     )
     print(createCollectionResult)
@@ -124,6 +138,7 @@ for d in data:
                 "updated_at": d[0],
                 "title": d[1],
                 "content": d[2],
+                "creator": d[3],
             },
         )
     except Exception as e:
--- a/backend/database-collection-schemas/createAnnouncementCollection.py
+++ b/backend/database-collection-schemas/createAnnouncementCollection.py
@@ -140,6 +140,8 @@ for d in data:
                 "content": d[2],
                 "creator": d[3],
             },
+            read=["*"],
+            write=["team:Admins"],
         )
     except Exception as e:
         print(e)
--- a/frontend/src/components/RestrictedArea.js
+++ b/frontend/src/components/RestrictedArea.js
@@ -4,7 +4,6 @@
 import React, { useEffect, useState } from "react";
 import appwriteApi from "../api/appwriteApi";
 import CenterFlexBox from "./CenterFlexBox";
-import ParagraphTypography from "./ParagraphTypography";
 import { Link } from "react-router-dom";
 import { activeTextColor } from "../assets/jss/colorPalette";
 import { arrayIntersection } from "../utils/utils";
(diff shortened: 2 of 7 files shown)


backend/database-collection-schemas/createAnnouncementCollection.py
```diff
@@ -105,14 +119,37 @@
...
+        "It's time where the Gimbleby sprites start to awake from their sleep. Look out and if you are lucky you catch one and get granted one wish!",
+        "robot",
+    ),
+    (
+        1647100404,
+        "Extreme Shopping Experience",
+        "Be one of the first to visit Xalax Ulcon at Heyla Dultherton's store today and win a first class augmented reality tour, the ultimate shopping experience.",
+        "robot",
+    ),
+    (1637100304, "Bella Balla", "Tam tam, tschuff, aba babba du", "robot"),
```
dominic-heil, line 152: Maybe we can replace this with something more generic?

dominic-heil (approved):
Runs on my machine. LGTM

backend/database-collection-schemas/createAnnouncementCollection.py
krisutofu: Yeah^^. I wanted to save my announcements somewhere. I can remove this since it is not relevant.

tuilagio (approved):

backend/database-collection-schemas/createAnnouncementCollection.py
```diff
@@ -94,6 +94,20 @@
                 "required": True,
                 "array": False,
             },
+            {
+                "label": "creator",
+                "key": "creator",
+                "type": "text",
```
steinkohl, line 100: rebuilt the whole script, so this will be obsolete - but included this fix.

backend/database-collection-schemas/createAnnouncementCollection.py
```diff
@@ -124,7 +138,10 @@
                 "updated_at": d[0],
                 "title": d[1],
                 "content": d[2],
+                "creator": d[3],
```
steinkohl, line 141: also removed dummy data creation from script, since they are not needed for production

krisutofu: That's okay.

backend/database-collection-schemas/createAnnouncementCollection.py
krisutofu: Okay,, so let me merge first so that you can overwrite it then.

tuilagio (approved):

tuilagio:
just tested on my machine. looks good

Floinko:
> Unfortunately, Appwrite seems to be not working correctly. The announcements initially added via Python cannot be found by using the JavaScript API, even when loading those documents explicitly via their document ID.

Sounds like a permission issue.
 
> One could also use Appwrite's API for listing database documents in a sorted way but the API seems to be unstable. It just was changed between v0.11 and v0.12 so a custom solution remains better in my opinion.

What do you mean by unstable? In the version, that we are using (0.11), you can filter and sort the documents...

krisutofu:
> Code looks good so far. I don't think that the custom sort function is necessary, but it also doesn't hurt to sort it locally (unless we have a LOT of announcements).

Note that this sort function previously was called in every rendering cycle and not only when updating the announcement array. I didn't bother changing that code and rather moved it instead of changing the API call ;-) .

